PR #63, opened 2024-07-24
Tinloof/2px button

freadbaker:
Implements 2px's Button component. The docs file is still just a placeholder.


apps/web/vibes/2px/components/button/index.tsx
```diff
@@ -1,64 +1,55 @@
...
+        primary:
+          'bg-foreground text-background hover:bg-background hover:text-foreground hover:border-2 hover:border-foreground',
+        secondary: 'bg-background text-foreground border-2 border-foreground hover:border-dashed',
+      },
+    },
+    defaultVariants: {
+      variant: 'primary',
+    },
+  }
+)
```
migueloller, line 22: I think that in an effort to minimize dependencies we're going to stay away from `class-variance-authority`. I know @apledger particularly despises this library. We're thinking that the same pattern can be applied doing something like this:

```tsx
<div className={clsx('tailwind classes', { primary: 'some more for primary', secondary: 'some more for secondary' }[variant])} />
```

migueloller (approved):

LHDi (approved):

PR #65, opened 2024-07-24
feat: Accordions component

freadbaker:
Implements 2px' Accordions component.


apps/web/vibes/2px/components.ts
```diff
@@ -3,6 +3,13 @@ import { lazy } from 'react'
 import { Components } from '@/vibes/schema'
 
 export const components = [
+  {
+    name: 'accordions',
+    dependencies: ['@radix-ui/react-accordion'],
+    registryDependencies: [],
+    files: ['./components/accordions/index.tsx'],
```
migueloller, line 10: I think we need to include the icons file in here as well since it's used by the Accordions component

apps/web/vibes/2px/components/accordions/index.tsx
migueloller, line 1: Given that we're not using state or effects here, can we get away with making this a shared component (i.e., drop `'use client'`?

migueloller (approved):
Looks good 👍🏻 

Got a couple of minor comments, nothing blocking.

apps/web/vibes/2px/components.ts
LHDi: I believe we missed this for all components. we'll make a PR to fix all components configs.

PR #66, opened 2024-07-24
feat: Tabs component

freadbaker:
Implements 2px' Tabs component.


apps/web/vibes/2px/components/tabs/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+    const activeItem = container.current.querySelector(`[data-state="active"]`)
+
+    if (!activeItem) return
+
+    const activeItemRect = activeItem.getBoundingClientRect()
+    const containerRect = container.current.getBoundingClientRect()
+
+    setActiveItemLeft(activeItemRect.left - containerRect.left + container.current.scrollLeft)
+    setActiveItemWidth(activeItemRect.width)
+  })
```
migueloller, line 54: Something interesting here is that we're calling `setState` and the `useEffect` in `usePollAnimationFrame` doesn't have a dependencies array. This means that these state changes could cause the component to go into an infinite render loop. I checked and it doesn't happen and I think it's because React is de-opting from the update if the state doesn't change to a different value by reference. Since the active item left and width are the same number, there isn't an issue. Maybe we want to set things up so that something like that isn't a problem. An easy way to avoid this is to put the callback in a ref that's set whenever `usePollAnimationFrame` is called and giving the `useEffect` an empty dependency array:

```ts
function usePollAnimationFrame(callback: (timestamp: number) => unknown) {
  const callbackRef = useRef(callback)

  callbackRef.current = callback

  useEffect(() => {
    let requestId: number

    function poll(timestamp: number) {
      requestId = requestAnimationFrame(poll)

      callbackRef.current(timestamp)
    }

    requestId = requestAnimationFrame(poll)

    return () => {
      cancelAnimationFrame(requestId)
    }
  }, [])
}
```

apps/web/vibes/2px/components/tabs/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+    requestId = requestAnimationFrame(poll)
+
+    return () => {
+      cancelAnimationFrame(requestId)
+    }
+  })
+}
+
+const Tabs = ({ className, tabs }: Props) => {
```
migueloller, line 36: This is a minor stylistic nit, but can we use function declarations for components? This will also make it consistent with examples:

```ts
export default function Tabs({ className,  tabs }: Props) {
```

migueloller (approved):
Looking great! I've got some comments but nothing that should block getting this shipped.

@apledger, I was looking at the preview on Vercel and the code for the example seems to be _just_ the example. Will the code showing the actual component source code come later once we auto-generate installation instructions?

apps/web/vibes/2px/components/tabs/index.tsx
LHDi: This wasn't needed in our case @migueloller, I just removed it. but surely what you proposed is much safer.

PR #76, opened 2024-07-25
feat: 2px radio button component

LHDi:


apps/web/vibes/2px/components/radio-button/index.tsx
```diff
@@ -0,0 +1,45 @@
+import clsx from 'clsx'
+
+interface RadioButtonProps extends Omit<React.HTMLProps<HTMLInputElement>, 'type' | 'hidden'> {
+  label: string
+}
+
+const RadioButton = ({ className, label, ...inputProps }: RadioButtonProps) => {
```
freadbaker, line 7: In another PR comment, Miguel had asked us to standardize our components as `export default function ...` instead of using `const`

apps/web/vibes/2px/examples/radio-button.tsx
```diff
@@ -0,0 +1,11 @@
+import RadioButton from '@/vibes/2px/components/radio-button'
+
+export default function Preview() {
+  return (
+    <div className="flex min-h-48 flex-col items-start justify-center gap-3 bg-background p-5 sm:min-h-64 sm:p-8 lg:min-h-80 lg:p-12">
+      <RadioButton label="Radio item" />
```
freadbaker, line 6: In this example, I'm able to select this radio button, but not unselect it later. I think it'd be nice to have that working both ways as an example.

freadbaker (commented):
Looks good to me, with the consideration that I think it'd be good to have the example fully working before merging.

apps/web/vibes/2px/examples/radio-button.tsx
LHDi: A radio button can't be unselected @freadbaker, checkboxes on the other hand can be unselected.

LHDi: Please check now @freadbaker, I made all of them the same name so they disable each other.

freadbaker: That's working great! Sorry for the confusion earlier.

apps/web/vibes/2px/components/radio-button/index.tsx
```diff
@@ -0,0 +1,44 @@
+import clsx from 'clsx'
+
+interface RadioButtonProps extends Omit<React.HTMLProps<HTMLInputElement>, 'type' | 'hidden'> {
```
freadbaker, line 3: @LHDi One thing I just recalled: in the [development guide](https://linear.app/tinloof/document/development-guide-bd8c365e62de), it says the components should not receive react element props. Do you think we can keep only the data props here?

LHDi: it's not passing react elements here @freadbaker

freadbaker: I had gotten this restriction wrong. I'm resolving this conversation, then.

PR #86, opened 2024-07-29
Feat: 2px dropdown component

LHDi:
This PR adds the 2px dropdown component: https://www.figma.com/design/JkWOwdQdlnrc04tUy3zlMK/Vibes---2px?node-id=5341-4817&m=dev


apps/web/vibes/2px/components/dropdown/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+        aria-label={placeholder}
+      >
+        {options.find(option => option.value === value)?.label || placeholder}
+
+        <ChevronDownIcon className="stroke-foreground" />
+      </button>
+
+      <div
+        className={clsx(
+          'fixed left-0 top-full z-10 mt-2 flex w-full flex-col items-start gap-1 border-2 border-foreground bg-background px-8 py-6 text-base font-medium leading-6 @lg:text-xl @lg:leading-9',
```
freadbaker, line 61: I believe the the fonts should be of `sm` and `2xl` sizes for small and big screens, respectively.

apps/web/vibes/2px/components/dropdown/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+  disabled,
+  setValue,
+  value,
+  className,
+}: DropdownProps) => {
+  const [isOpen, setIsOpen] = useState(false)
+
+  return (
+    <div className={clsx('relative w-full font-body text-foreground @container', className)}>
+      <button
```
freadbaker, line 37: I think it'd be good to fix the height as 3.75rem and 5rem for small and big screens, respectively.

apps/web/vibes/2px/components/dropdown/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+          }
+        )}
+        disabled={disabled}
+        onClick={() => setIsOpen(!isOpen)}
+        type="button"
+        aria-label={placeholder}
+      >
+        {options.find(option => option.value === value)?.label || placeholder}
+
+        <ChevronDownIcon className="stroke-foreground" />
```
freadbaker, line 56: The chevron icon should decrease in size for smaller screens.

apps/web/vibes/2px/components/dropdown/index.tsx
freadbaker: Also, the larger screens should have a `letter-spacing: -0.0175rem;` according to the design.

freadbaker (changes requested):
Looking good, just a few comments mostly on typography.

apps/web/vibes/2px/components/dropdown/index.tsx
LHDi: Actually it's the same size on both small and large screens

apps/web/vibes/2px/components/dropdown/index.tsx
LHDi: It's actually not @freadbaker , let's say someone uses this with another vibe where that line-height is bigger. having a fixed height is bad in that case.
As a rule of thumb if you can keep the height or width responsive do it.

apps/web/vibes/2px/components/dropdown/index.tsx
LHDi: Done 🙏

apps/web/vibes/2px/components/dropdown/index.tsx
freadbaker: @LHDi In [the design file](https://www.figma.com/design/JkWOwdQdlnrc04tUy3zlMK/Vibes---2px?node-id=5341-4817&m=dev) it seems like the icon is a 1rem x 1rem square for smaller screens while 1.5rem x 1.5rem for bigger screens.

Am I missing something?

LHDi: You are right! and I am right! lol

I've been following the icon sizes of the open version.
image

I'll follow yours, it makes more sense 🫡

PR #138, opened 2024-08-15
Feat: 2px date picker component

LHDi:


apps/web/vibes/2px/components/date-picker/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const [calendarOpen, setCalendarOpen] = useState(false)
+  const calendarRef = useRef<HTMLDivElement>(null)
+
+  useOnClickOutside(calendarRef, () => setCalendarOpen(false))
+
+  const handleCalendarSelect = (date?: Date) => {
+    const formattedDate = date?.toISOString()
+    ;(onChange || setSelectedDate)(
+      formattedDate
+        ? parseDate(parseAbsoluteToLocal(formattedDate).toString().split('T')[0])
```
freadbaker, line 37: Whenever I select a date in the calendar, the day before gets painted, although the right date is inserted in the input. 

I think this might be caused by the timezone difference. 

image

apps/web/vibes/2px/components/icons/CalendarIcon.tsx
```diff
@@ -10,9 +10,10 @@ export const CalendarIcon: React.FC<React.ComponentPropsWithoutRef<'svg'>> = ({
       viewBox="0 0 16 16"
       fill="none"
       className={className}
+      stroke="black"
```
freadbaker, line 13: Should we make the stroke be of currentColor instead?

apps/web/vibes/2px/examples/date-picker.tsx
```diff
@@ -0,0 +1,12 @@
+import DatePicker from '@/vibes/2px/components/date-picker'
+
+export default function Preview() {
+  return (
+    <div className="flex max-w-lg flex-col items-center justify-center gap-5 bg-background p-5 sm:min-h-64 sm:p-8 lg:min-h-[500px] lg:p-12">
+      <DatePicker />
+      <DatePicker disabled />
```
freadbaker, line 7: Since the popup is rendered at the top, I suggest starting with the disabled version at the top to give some more room for the enabled date-pickers' calendars.

freadbaker (changes requested):
Great work! There's one major concern with the calendar date showing wrong in my timezone, but all other comments are minor.

PR #183, opened 2024-09-11
soul: ESLint fixes

Samantha2233:


apps/web/vibes/soul/components/blog-post-card/index.tsx
```diff
@@ -36,16 +36,18 @@ export const BlogPostCard = function BlogPostCard({
...
+          <Image
+            src={image.src}
+            height={349}
+            width={466}
+            sizes="(max-width: 768px) 100vw, (max-width: 1200px) 50vw, 33vw"
+            alt={image.altText}
+            className="bg-contrast-100 transition-transform duration-500 ease-out group-hover:scale-105"
+          />
+        </div>
+      )}
```
migueloller, line 50: Should there be a fallback when there's no image source?

apps/web/vibes/soul/components/blog-post-card/index.tsx
```diff
@@ -61,8 +63,8 @@ export const BlogPostCard = function BlogPostCard({
             })}
           </time>
         )}
-        {date && author && <span className="after:mx-2 after:content-['•']" />}
-        {author && author}
+        {date && author !== undefined && <span className="after:mx-2 after:content-['•']" />}
+        {author !== undefined && <span>author</span>}
```
migueloller, line 67: I think for these we want to check `date != null && author != null`. Also, seems like we're missing angled brackets around `author`, should be `author != null && <span>{author}</span>`

apps/web/vibes/soul/components/button/index.tsx
```diff
@@ -41,7 +41,7 @@ export const Button = function Button({
         'after:absolute after:inset-0 after:z-0 after:h-full after:w-full after:rounded-full',
         'after:ease-[cubic-bezier(0.075,0.365,0.000,0.995)] after:transition-[opacity,transform] after:duration-500',
         'after:-translate-x-[110%]',
-        !loading && !disabled && 'hover:after:translate-x-0',
+        !Boolean(loading) && !Boolean(disabled) && 'hover:after:translate-x-0',
```
migueloller, line 44: Instead of coercing these values, I think we should provide default values so that they're always a boolean. Alternatively, we can do `loading !== true` or `disabled !== true`. I personally prefer having the values always have a boolean as it makes things easier to follow

apps/web/vibes/soul/components/carousel/index.tsx
```diff
@@ -59,7 +59,7 @@ export const Carousel = ({
       className={clsx('mx-auto flex w-full max-w-screen-2xl flex-col gap-6 @container', className)}
     >
       {/* Carousel Content */}
-      {children && (
+      {Boolean(children) && (
```
migueloller, line 62: I would avoid coercing values using type constructors and instead check `children != null`

migueloller: Why do we check for `children` here anyway? Is there a situation where we expect to receive a falsy ReactNode and don't want to render anything here? Maybe we want a fallback in that scenario?

apps/web/vibes/soul/components/checkbox/index.tsx
```diff
@@ -33,7 +33,7 @@ export const Checkbox = function Checkbox({
         </RadixCheckbox.Indicator>
       </RadixCheckbox.Root>
 
-      {label && (
+      {label !== undefined && (
```
migueloller, line 36: I'd avoid explicit comparisons against `undefined` and instead check `== null`

apps/web/vibes/soul/components/compare-card/index.tsx
```diff
@@ -52,10 +52,10 @@ export const CompareCard = function CompareCard({
       href={href}
     >
       <div className="relative aspect-[5/6] overflow-hidden rounded-xl">
-        {badge && (
+        {badge !== undefined && (
```
migueloller, line 55:
```suggestion
        {badge != null && (
```

apps/web/vibes/soul/components/compare-card/index.tsx
```diff
@@ -68,28 +68,30 @@ export const CompareCard = function CompareCard({
       <div className="mb-2 flex flex-col gap-1">
         <h3 className="flex flex-col flex-wrap justify-between gap-1 text-sm font-semibold @sm:pt-3 @4xl:flex-row">
           {name && <span className="line-clamp-2">{name}</span>}
-          {subtitle && <span className="font-normal text-contrast-400">{subtitle}</span>}
+          {subtitle !== undefined && (
```
migueloller, line 71:
```suggestion
          {subtitle != null && (
```

apps/web/vibes/soul/components/compare-card/index.tsx
```diff
@@ -68,28 +68,30 @@ export const CompareCard = function CompareCard({
...
+            <span className="font-normal text-contrast-400">{subtitle}</span>
+          )}
         </h3>
-        {price && <Price price={price} />}
+        {price !== undefined && <Price price={price} />}
```
migueloller, line 75:
```suggestion
        {price !=  null && <Price price={price} />}
```

apps/web/vibes/soul/components/compare-card/index.tsx
```diff
@@ -68,28 +68,30 @@ export const CompareCard = function CompareCard({
...
       </div>
       <Button className="mb-8 w-full">Add to Cart</Button>
       <hr className="mb-4" />
 
-      {description && (
+      {description !== undefined && (
```
migueloller, line 80:
```suggestion
      {description != null && (
```

apps/web/vibes/soul/components/compare-card/index.tsx
```diff
@@ -68,28 +68,30 @@ export const CompareCard = function CompareCard({
...
-      {rating && (
+      {rating != null && (
         <>
           <Label className="mb-3">Rating</Label>
           <Rating rating={rating} className="mb-8" />
           <hr className="mb-4" />
         </>
       )}
-      {availability && (
+      {availability !== undefined && (
```
migueloller, line 94:
```suggestion
      {availability != null && (
```

apps/web/vibes/soul/components/compare-drawer/product-chip.tsx
```diff
@@ -15,22 +15,21 @@ interface Props {
...
         })
-      }
+      }}
       className={clsx(
         'group flex items-center gap-3 whitespace-nowrap rounded-xl border border-contrast-100 bg-background font-semibold transition-all duration-150 hover:bg-contrast-100',
         'ring-primary focus:outline-0 focus:ring-2'
       )}
     >
-      {product.image?.src && (
+      {product.image?.src !== undefined && (
```
migueloller, line 29:
```suggestion
      {product.image?.src != null && (
```

apps/web/vibes/soul/components/discount/index.tsx
```diff
@@ -67,21 +67,22 @@ export const Discount = function Discount({ backgroundImage, discounts }: Props)
...
       >
         <X className="h-6 w-6" />
       </button>
 
-      <div
+      <button
         onClick={() => {
           if (isSpun) {
-            copy()
+            void copy()
```
migueloller, line 81: Is this `void` keyword necessary here?

apps/web/vibes/soul/components/discount/index.tsx
```diff
@@ -98,7 +99,7 @@ export const Discount = function Discount({ backgroundImage, discounts }: Props)
               className="absolute -top-8 left-0 w-full transition-all [transition-duration:5000ms] [transition-timing-function:cubic-bezier(0.285,-0.125,0.050,1.130)]"
               style={{
                 transform: spin
-                  ? `translateY(calc(-100% + ${discounts.length * 33}px))`
+                  ? `translateY(calc(-100% + ${String(discounts.length * 33)}px))`
```
migueloller, line 102: Is it really necessary to coerce a number to a string to interpolate? If there's a rule forcing that then it might a silly one.

apps/web/vibes/soul/components/dropdown/index.tsx
```diff
@@ -61,7 +61,9 @@ export const Dropdown = function Dropdown({
                   'text-foreground': selectedItem === item,
                 }
               )}
-              onSelect={() => setSelectedItem(item)}
+              onSelect={() => {
+                setSelectedItem(item)
+              }}
```
migueloller, line 66: What particular rule is asking for this change? It seems like this shouldn't matter

apps/web/vibes/soul/components/favorite/heart.tsx
```diff
@@ -1,10 +1,12 @@
+import React from 'react'
+
```
migueloller, line 2: The React import shouldn't be necessary here

migueloller (changes requested):
A couple of things I'm noticing here:
- If there's a rule that requires coercing a value to a string before interpolating, we should probably disable that, it's a silly rule
- When comparing optional values, let's opt for `==` or `!=` with `null` instead of `undefined`
- Some of the changes made seem to be changing the behavior of components so I think we should double check that we have an appropriate fallback. For example, when avoiding to render an image if the image source isn't provided
- We're using indices as React keys and this can be problematic. Maybe we should figure out if the prop interfaces need to change so that we have a key for lists
- We should avoid coercing values using type constructors like `Boolean` or `String` since it does implicit conversations and should try to be explicit as much as we can
- We seem to have a rule that's making us change all arrow functions to avoid returning a value. This rule seems silly, let's disable it.

All other changes look great to me 👍🏻

apps/web/vibes/soul/components/discount/index.tsx
Samantha2233: The `copy()` function is an async function that returns a Promise. ESLint's `@typescript-eslint/no-floating-promises` was warning about the unhandled promise since it can lead to unhandled rejections. Using `void` here indicates that we're aware this function returns a Promise, but we're intentionally not handling it - silencing the linter warning without changing the behavior of the code.

Let me know if you think there is a better way to handle this though! I'm open.

apps/web/vibes/soul/components/card/index.tsx
```diff
@@ -5,7 +5,8 @@ import { ComponentPropsWithoutRef } from 'react'
 import { clsx } from 'clsx'
 import { ArrowUpRight } from 'lucide-react'
 
-export type CardProps = {
+export interface CardProps {
```
migueloller, line 8: Seems a bit weird to require an `id` prop if the component doesn't use it

apps/web/vibes/soul/components/compare-card/index.tsx
```diff
@@ -68,28 +68,30 @@ export const CompareCard = function CompareCard({
       <div className="mb-2 flex flex-col gap-1">
         <h3 className="flex flex-col flex-wrap justify-between gap-1 text-sm font-semibold @sm:pt-3 @4xl:flex-row">
           {name && <span className="line-clamp-2">{name}</span>}
-          {subtitle && <span className="font-normal text-contrast-400">{subtitle}</span>}
+          {subtitle != null && (
+            <span className="font-normal text-contrast-400">{subtitle}</span>
+          )}
```
migueloller, line 73: Just want to double-check that for some of these the intent wasn't to hide the entire element if the string was empty? i.e., do we actually want to do `subtitle && ...` but ignore the ESLint rule? Or perhaps be explicit and do `subtitle != null && subtitle !== '' && ...`. Same goes for other components were if the prop is an empty string we don't want to render anything

apps/web/vibes/soul/components/discount/index.tsx
migueloller: I would usually opt for handling the promise with a `.catch` or `.then` but I think this is a reasonable approach 👍🏻

apps/web/vibes/soul/components/featured-blog-post-list/index.tsx
```diff
@@ -28,9 +28,9 @@ export const FeaturedBlogPostList = function FeaturedBlogPostList({
         <h2 className="mb-2 font-heading text-3xl font-semibold leading-none text-foreground @4xl:text-6xl @4xl:font-medium">
           {title}
         </h2>
-        {description && <p className="max-w-md text-foreground">{description}</p>}
+        {description != null && <p className="max-w-md text-foreground">{description}</p>}
```
migueloller, line 31: See my other comments on not rendering elements when the string is empty

apps/web/vibes/soul/components/header/index.tsx
```diff
@@ -15,18 +15,23 @@ import {
...
+import { HamburgerMenuButton } from '@/vibes/soul/components/header/hamburger-menu-button'
+
 interface Image {
   src?: string
   altText: string
 }
 
-interface Links {
+export interface Links {
+  id: string
```
migueloller, line 26: It seems weird to have `id` on `Links` if it isn't an array item

apps/web/vibes/soul/components/input/index.tsx
```diff
@@ -18,9 +18,9 @@ export const Input = forwardRef(function Input(
 ) {
   return (
     <div className={clsx('w-full shrink-0', className)}>
-      {label && <Label className="mb-2 block text-foreground">{label}</Label>}
+      {label != null && <Label className="mb-2 block text-foreground">{label}</Label>}
       <div className="relative overflow-hidden rounded-lg border border-contrast-100 bg-background transition-colors duration-200 focus-within:border-foreground focus:outline-none">
-        {prepend && (
+        {prepend != null && (
```
migueloller, line 23: Do we want to render these elements when the values are `''`?

apps/web/vibes/soul/examples/card.tsx
```diff
@@ -4,6 +4,7 @@ export default function Preview() {
   return (
     <div className="flex min-h-48 items-center justify-center p-5 @container">
       <Card
+        id="1"
```
migueloller, line 7: This prop shouldn't be in `CardProps`, only when there's a list. In those cases, it's likely that it's not a "card" that will have an ID but the thing being mapped on. For example, maybe there's a mapping over products so the product ID can be used as a key for each card

migueloller (approved):

PR #193, opened 2024-10-04
add no-plusplus, radix, no-nested-ternary, & style-prop-object rules

jorgemoya:
Adds some additional rules that we use in the Catalyst repo that error when copy pasting components over.

### DIFF
--- a/apps/web/.eslintrc.json
+++ b/apps/web/.eslintrc.json
@@ -15,6 +15,9 @@
   "parserOptions": {
     "project": "./tsconfig.json"
   },
+  "plugins": [
+    "react"
+  ],
   "rules": {
     "@typescript-eslint/no-confusing-void-expression": [
       "error",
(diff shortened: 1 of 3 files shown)


apps/web/.eslintrc.json
```diff
@@ -15,6 +15,9 @@
   "parserOptions": {
     "project": "./tsconfig.json"
   },
+  "plugins": [
+    "react"
+  ],
```
migueloller, line 20: I thought `next/core-web-vitals` already included the `eslint-plugin-react`. Did you try adding the rule without introducing the plugin here? If rules weren't included, maybe we want to add `plugin:react/recommended` to `extends` instead of adding the plugin here that way we enforce recommended React patterns, like the rules of hooks.

migueloller (approved):
Looks good! Just have a comment/question about how to use `eslint-plugin-react` and whether it's already included with the Next.js plugin.

apps/web/.eslintrc.json
jorgemoya: I didn't try, let me check it out. 👍

jorgemoya: Seems like it is included like you suggested, so I removed the standalone plugin install.

migueloller (approved):

PR #229, opened 2024-10-17
chore: Isolate Soul Vibe

Samantha2233:


apps/web/vibes/index.ts
```diff
@@ -1,3 +1,3 @@
-export { default as eclipse } from './eclipse'
+// export { default as eclipse } from './eclipse'
```
migueloller: Should we go ahead and delete these commented out lines? The other branches should have them anyway so we don't need to keep them around.

apps/web/vibes/schema.ts
```diff
@@ -110,14 +111,18 @@ export const pageMetaSchema = (vibe: Vibe) => {
...
-        z.object(
-          Object.fromEntries(
-            vibe.brands.map(brand => [
-              brand.name,
-              z.enum([component.name, ...rest.map(entry => entry.name)]),
-            ])
+        z
+          .object(
+            Object.fromEntries(
+              vibe.brands.length > 0
```
migueloller: Do we ever want to support a VIBE with 0 brands?

apps/web/vibes/schema.ts
```diff
@@ -14,7 +14,7 @@ export type Component = {
   >
 }
 
-export type Components = NonEmptyArray<Component>
+export type Components = Array<Component>
```
migueloller: Do we want to support empty arrays for values? It seems that a VIBE would include at least one of these

migueloller: Ok, I'm seeing now that we're using this for "coming soon" VIBES. Maybe we want to have a separate type for that so we're not making the main type have to change to satisfy a "coming soon" VIBE? In particular since we don't need navigation, components, brands, tags, etc.

migueloller (approved):
This looks good. I've just got some minor comments about keeping our types strict even if we're adding "coming soon" VIBES.

What's the plan with the other ones, to have branches we'll merge later? If so, I'd like to start protecting `main` since `main` won't be as WIP anymore

PR #244, opened 2024-10-18
chore: split components file into primitives and sections

Samantha2233:


jamesqquick:
This looks good to me except there's an error running locally.

``` Functions cannot be passed directly to Client Components unless you explicitly expose it by marking it with "use server". Or maybe you meant to call this function rather than return it.```

I still haven't been able to figure this out but it seems to happen when rendering `content` variable generated from the docs Markdown

jamesqquick (approved):

PR #248, opened 2024-10-22
Refactor Soul accordion primitive

migueloller:
This PR refactors the accordion primitive
- Renames from "accordions" to "accordion"
- Moves from `accordions/index.tsx` to `accordion.tsx`
- Mimics Radix UI API
- Adds JSDoc to new props
- Consolidates component and exports for easier visual parsing


apps/web/vibes/soul/primitives/accordion.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export const Root = AccordionsPrimitive.Root
+
+type Props = React.ComponentPropsWithoutRef<typeof AccordionsPrimitive.Item> & {
+  /**
+   * The title of the accordion item. It will be added to the accordion's header.
+   */
+  title: string
+
+  children: React.ReactNode
```
jorgemoya, line 13: `typeof AccordionsPrimitive.Item` should already include `children` no?

jorgemoya (approved):

apps/web/vibes/soul/primitives/accordion.tsx
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export function Item({ title, children, ...rest }: Props) {
```
jorgemoya, line 16: Should we include a `displayName`?

migueloller: Won't the display name be `Item` since this is a function declaration? I've only seen explicit `displayName` necessary when using `forwardRef` with an anonymous function resulting in an unnamed component

jorgemoya: Ah yes, that makes sense.

PR #249, opened 2024-10-23
Andrew/component style fixes

andrewreifman:


apps/web/next.config.mjs
```diff
@@ -11,7 +11,7 @@ const nextConfig = {
   pageExtensions: ['js', 'jsx', 'mdx', 'ts', 'tsx'],
   // Optionally, add any other Next.js config below
   images: {
-    domains: ['images.unsplash.com', 'rstr.in', 'images.pexels.com'],
+    domains: ['images.unsplash.com', 'rstr.in', 'images.pexels.com', 'storage.googleapis.com'],
```
migueloller, line 14: I think this might be too permissive. We should move to `remotePatterns` and specify the Makeswift bucket here specifically. More here: https://nextjs.org/docs/app/api-reference/components/image#remotepatterns

apps/web/vibes/soul/primitives/animated-link/index.tsx
```diff
@@ -12,8 +11,15 @@ export interface AnimatedLinkProps {
...
   return (
-    <Link href={href} className={clsx('animated-link animated-underline', className)}>
-      <span>{label}</span>
+    <Link
+      href={href}
+      className={clsx(
+        'origin-left pb-0.5 font-semibold transition-[background-size] duration-500 [background:linear-gradient(0deg,hsl(var(--primary)),hsl(var(--primary)))_no-repeat_left_bottom_/_0_var(--bg-h)] hover:bg-[size:100%_var(--bg-h)] focus:outline-none focus-visible:bg-[size:100%_var(--bg-h)]',
+        className
+      )}
+      style={{ '--bg-h': '2px' } as CSSProperties}
```
migueloller, line 20: Can't we use a computed property in the Tailwind classes? How come we pass it via an inline style if it doesn't depend on state?

apps/web/vibes/soul/primitives/navigation/index.tsx
```diff
@@ -359,9 +352,69 @@ export const Header = forwardRef(function Header(
               ))}
             </div>
             <div className="flex w-full flex-col gap-5 p-5">
-              <span className="font-mono text-[13px] uppercase">Products</span>
+              <span className="font-mono text-sm uppercase">Products</span>
               <div className="grid w-fit grid-cols-2 gap-5 @xl:grid-cols-4 @2xl:grid-cols-2 @4xl:grid-cols-4">
-                {Array.from({ length: 4 }).map((_, i) => (
+                <ProductCard
+                  key="1"
```
migueloller, line 358: Don't need these keys anymore since we're not using an array

migueloller (changes requested):
This looks good! Just have a couple of comments and we can ship this

apps/web/vibes/soul/primitives/animated-link/index.tsx
andrewreifman: Just went ahead and removed the css variable since it's only used in this component in very few places.

andrewreifman:
@migueloller Ready to merge if you are

migueloller (approved):

PR #250, opened 2024-10-23
feat: adds Calendar primitive

jorgemoya:
Adds a `calendar` primitive that will be used for the `DatePicker` form component.

Screenshot 2024-10-28 at 3 04 24 PM

### DIFF
--- a/apps/web/vibes/soul/primitives/calendar.tsx
+++ b/apps/web/vibes/soul/primitives/calendar.tsx
@@ -1,0 +1,44 @@
+import { ComponentPropsWithoutRef } from 'react'
+import { DayPicker } from 'react-day-picker'
+
+import clsx from 'clsx'
+import { ChevronLeftIcon } from 'lucide-react'
+
+export function Calendar({
+  className,
+  classNames,
+  ...props
+}: ComponentPropsWithoutRef<typeof DayPicker>) {
+  return (
+    <DayPicker
+      className={clsx('w-[304px] rounded-lg border border-contrast-100 p-3', className)}
+      classNames={{
+        months: 'relative',
+        month_caption: 'flex justify-center w-full font-medium',
+        nav: 'absolute flex justify-between w-full',
+        button_next:
+          'rotate-180 rounded-full focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-foreground',
+        button_previous:
+          'rounded-full focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-foreground',
+        weekdays: 'flex py-0.5',
+        weekday: 'flex h-10 w-10 items-center justify-center text-xs font-medium',
+        week: 'flex py-0.5',
+        day: 'h-10 w-10 flex text-xs font-medium group',
+        day_button:
+          'h-full w-full flex items-center justify-center rounded-full group-data-[selected=true]:bg-primary group-data-[selected=true]/middle:bg-primary-highlight hover:border hover:border-contrast-200 focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-foreground disabled:hover:border-none',
+        disabled: 'text-contrast-300',
+        outside: 'text-contrast-300',
+        range_start: 'bg-gradient-to-l from-primary-highlight',
+        range_middle: 'bg-primary-highlight group/middle',
+        range_end: 'bg-gradient-to-r from-primary-highlight',
+        ...classNames,
+      }}
+      components={{
+        Chevron: () => <ChevronLeftIcon className="h-5 w-5" strokeWidth={1} />,
+      }}
+      {...props}
+    />
+  )
+}
+
+Calendar.displayName = 'Calendar'
--- a/apps/web/vibes/soul/primitives/calendar.tsx
+++ b/apps/web/vibes/soul/primitives/calendar.tsx
@@ -45,5 +45,3 @@ export function Calendar({
     />
   )
 }
-
-Calendar.displayName = 'Calendar'
--- a/apps/web/vibes/soul/primitives/calendar.tsx
+++ b/apps/web/vibes/soul/primitives/calendar.tsx
@@ -17,15 +17,15 @@ export function Calendar({
       )}
       classNames={{
         months: 'relative',
-        month_caption: 'flex justify-center w-full font-medium',
+        month_caption: 'flex justify-center w-full font-medium pb-px',
         nav: 'absolute flex justify-between w-full',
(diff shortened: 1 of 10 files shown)


apps/web/vibes/soul/form/date-picker/index.tsx
migueloller, line 1: Was it intentional to commit this empty file?

apps/web/vibes/soul/primitives/calendar.tsx
```diff
@@ -0,0 +1,49 @@
...
+  className,
+  classNames,
+  ...props
+}: ComponentPropsWithoutRef<typeof DayPicker>) {
+  return (
+    <DayPicker
+      className={clsx(
+        'box-content w-[280px] rounded-lg border border-contrast-100 bg-background p-3',
+        className
+      )}
```
migueloller, line 17: This is a place where we'd definitely want to be using `twMerge`

apps/web/vibes/soul/primitives/calendar.tsx
```diff
@@ -0,0 +1,49 @@
...
+      }}
+      components={{
+        Chevron: () => <ChevronLeftIcon className="h-5 w-5" strokeWidth={1} />,
+      }}
+      {...props}
+    />
+  )
+}
+
+Calendar.displayName = 'Calendar'
```
migueloller, line 49: I think this display name is redundant since the function name is already `Calendar`

apps/web/vibes/soul/primitives/calendar.tsx
```diff
@@ -0,0 +1,49 @@
...
+          'h-full w-full flex items-center justify-center rounded-full group-data-[selected=true]:bg-primary group-data-[selected=true]/middle:bg-transparent hover:border hover:border-contrast-200 focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-foreground disabled:hover:border-none',
+        disabled: 'text-contrast-300',
+        outside: 'text-contrast-300',
+        range_start: 'bg-gradient-to-l from-primary-highlight',
+        range_middle: 'bg-primary-highlight group/middle',
+        range_end: 'bg-gradient-to-r from-primary-highlight',
+        ...classNames,
+      }}
+      components={{
+        Chevron: () => <ChevronLeftIcon className="h-5 w-5" strokeWidth={1} />,
```
migueloller, line 42: Is it a concern that on every render the `Chevron` component will have a different identity?

apps/web/vibes/soul/primitives/calendar.tsx
```diff
@@ -0,0 +1,49 @@
...
+        week: 'flex',
+        day: 'h-10 w-10 flex text-xs font-medium group p-0',
+        day_button:
+          'h-full w-full flex items-center justify-center rounded-full group-data-[selected=true]:bg-primary group-data-[selected=true]/middle:bg-transparent hover:border hover:border-contrast-200 focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-foreground disabled:hover:border-none',
+        disabled: 'text-contrast-300',
+        outside: 'text-contrast-300',
+        range_start: 'bg-gradient-to-l from-primary-highlight',
+        range_middle: 'bg-primary-highlight group/middle',
+        range_end: 'bg-gradient-to-r from-primary-highlight',
+        ...classNames,
```
migueloller, line 39: We might not want to expose all classes here. WDYT?

apps/web/vibes/soul/form/date-picker/index.tsx
jorgemoya: No, will remove!

apps/web/vibes/soul/primitives/calendar.tsx
jorgemoya: As a primitive I think they should have full control of how they want to override styles over the default values, but not set on keeping this. Are you saying I should remove this entirely?

apps/web/vibes/soul/primitives/calendar.tsx
jorgemoya: Seems to be the [way to implement it](https://daypicker.dev/guides/custom-components#implementing-a-custom-component). What would you suggest?

migueloller: If this is how they document it, then I have no issues here. They might be calling it as a "render function" instead of using it as a component (I hope...)

migueloller: Ugh, looks like they don't: https://github.com/gpbl/react-day-picker/blob/903745cb6c1332818ef558729c0c598a18563767/src/components/Nav.tsx#L44

The solution here would be to define the component outside of `Calendar` so that it's not a new instance on each render. I'm surprised this library has this React antipattern in their docs like this.

migueloller: I guess it's not a big deal if the chevron remounts but it's just bad practice

migueloller: Here's somebody running into the problem: https://github.com/gpbl/react-day-picker/issues/839#issuecomment-444150953

I say let's just pull out the `components` object outside the component as a top level `const`

apps/web/vibes/soul/primitives/calendar.tsx
migueloller: I was just wondering with VIBES specifically if somebody wanted to change how their calendar looked they could go in and customize the source code. This isn't a big deal and we can always come back later and remove it. I think this is where I would align with Alan and I wouldn't expose all classes since this component isn't a headless component and is meant to encapsulate all styles in a cohesive way.

Feel free to leave as-is for now

jorgemoya: I will remove for now, pretty simple to add back if needed.

apps/web/vibes/soul/primitives/calendar.tsx
```diff
@@ -0,0 +1,48 @@
...
+import clsx from 'clsx'
+import { ChevronLeftIcon } from 'lucide-react'
+
+const components = {
+  Chevron: () => <ChevronLeftIcon className="h-5 w-5" strokeWidth={1} />,
+}
+
+export function Calendar({
+  className,
+  classNames,
```
migueloller, line 13: This prop is now unused. Instead of spreading all props we might want to only expose a few? As it stands right now if `classNames` is passed, TypeScript won't complain but it will be ignored.

migueloller (approved):

PR #255, opened 2024-10-29
fix: use radio group in ToggleGroup

jorgemoya:
Replaces Radix's toggle group for radio group since it adds hidden inputs inside forms.


migueloller (approved):
LGTM 👍🏻

jorgemoya:
Seems like there is an use case of a toggle group that is multiple choice and that is not inside a form, for filters in PLP. In which case maybe there needs to be two separate components.

migueloller:
> Seems like there is an use case of a toggle group that is multiple choice and that is not inside a form, for filters in PLP. In which case maybe there needs to be two separate components.

We discussed in Slack. I think we'll want the PLP toggle group to be a form field as well but it's not a requirement for 11/11 so for now do whatever's fastest 👍🏻

jorgemoya:
Closing because we need this for filters, creating a new one for forms using RadioGroupPrimitive.

PR #258, opened 2024-10-30
feat: add missing searchHref prop in Navigation

jorgemoya:
Just like `accountHref` and `cartHref`, we need to allow passing in a custom search href the form submits too.

### DIFF
--- a/apps/web/vibes/soul/examples/pages/blog/index.tsx
+++ b/apps/web/vibes/soul/examples/pages/blog/index.tsx
@@ -67,6 +67,7 @@ export default function Preview() {
         accountHref="#"
         locales={locales}
         activeLocale="EN"
+        searchHref="#"
       />
 
       <FeaturedBlogPostList
--- a/apps/web/vibes/soul/examples/pages/blog/post/electric.tsx
+++ b/apps/web/vibes/soul/examples/pages/blog/post/electric.tsx
@@ -73,6 +73,7 @@ export default function Preview() {
         accountHref="#"
         locales={locales}
         activeLocale="EN"
+        searchHref="#"
       />
 
       <BlogPostContent
--- a/apps/web/vibes/soul/examples/pages/blog/post/luxury.tsx
+++ b/apps/web/vibes/soul/examples/pages/blog/post/luxury.tsx
@@ -73,6 +73,7 @@ export default function Preview() {
         accountHref="#"
         locales={locales}
         activeLocale="EN"
+        searchHref="#"
       />
 
       <BlogPostContent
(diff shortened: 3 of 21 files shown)


apps/web/vibes/soul/sections/header/index.tsx
```diff
@@ -147,7 +158,7 @@ export const Header = forwardRef(function Header(
     const searchTerm = searchInputRef.current?.value.trim() ?? ''
 
     if (searchTerm.length > 0) {
-      router.push(`/search?term=${encodeURIComponent(searchTerm)}`)
+      router.push(`/${searchHref}?term=${encodeURIComponent(searchTerm)}`)
```
migueloller, line 161: Minor nit on the naming here. If we're naming it "href" it should probably be the full  relative or absolute URL, no? If so, we don't want to pre-pend the leading `/`? Also, if the provided URL already has search params then hardcoding `?` might give unexpected results.

migueloller: Also, it seems we pass `accountHref` to an `href` attribute but for this one we do some more custom stuff. Might not be relevant for this PR but perhaps we want to handle this with a different approach?

migueloller (changes requested):

apps/web/vibes/soul/sections/header/index.tsx
jorgemoya: Total oversight on my part, href should include `/` like you mention.

I think this is a simple form action that would use the new form component in Next 15. All we want to do is trigger a redirect to `/search?term=blah`. I don't think we should care about passing in any other search params with this redirect, it's not commonly a thing afaik.

migueloller:
> As a side note, should the search term param also be a prop? 🤔

Generally everything should be configurable, so yeah. Some might use `query`, others `q`, etc.

apps/web/vibes/soul/sections/header/index.tsx
migueloller: Yeah. Do you want to take care of using the new form component in this PR or leave is as a todo for a future PR? My guess is we'll come back to this header component and do another pass before 11/11.

migueloller (approved):

apps/web/vibes/soul/sections/header/index.tsx
jorgemoya: Lets solve in a different PR for a different time.

jorgemoya:
@migueloller Added searchHref as an optional prop.

apledger:
> > As a side note, should the search term param also be a prop? 🤔
> 
> Generally everything should be configurable, so yeah. Some might use `query`, others `q`, etc.

Agreed. That would be consistent with the patterns we have in the PLP

migueloller (approved):

PR #260, opened 2024-10-31
feat: add search to navigation

jorgemoya:
+ Adds all props necessary to allow fetching results when search is typed in the navigation component.
+ Search is debounced and only triggers after a couple of seconds.
+ Split search results to its own `SearchResults` component.
+ Add loading state to search button when search is pending
+ Updated the `<Spinner />` component to have `xs` size needed for this specific button.
+ Add translation labels for empty state.

Screenshot 2024-10-31 at 1 19 58 PM


apps/web/package.json
```diff
@@ -57,6 +57,7 @@
     "gray-matter": "^4.0.3",
     "gzip-size": "^7.0.0",
     "jiti": "^1.21.3",
+    "lodash.debounce": "^4.0.8",
```
jorgemoya, line 60: Using this library for debounce but open to suggestions.

apps/web/vibes/soul/primitives/spinner/index.tsx
```diff
@@ -1,7 +1,7 @@
 import { clsx } from 'clsx'
 
 interface Props {
-  size?: 'sm' | 'md' | 'lg'
+  size?: 'xs' | 'sm' | 'md' | 'lg'
```
jorgemoya, line 4: I could avoid this by increasing size of search icon in search form.

migueloller (approved):

PR #278, opened 2024-11-06
Header refactor

apledger:


apps/web/vibes/soul/data/locales.ts
```diff
@@ -0,0 +1,6 @@
+export const locales = [
+  { id: 'US', label: 'United States' },
+  { id: 'FR', label: 'France' },
+  { id: 'DE', label: 'Denmark' },
+  { id: 'IT', label: 'Italy' },
```
migueloller, line 5: These are invalid locale identifiers. Locale identifiers require a language code while the country code is optional. Maybe you meant to use `en`, `fr`, `de`, and `it`?

apps/web/vibes/soul/examples/sections/header-section/electric.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <>
+      <HeaderSection
+        banner={{
+          children: (
+            <>
+              Get <strong>15% off</strong> and free shipping with discount code{' '}
+              <strong>&quot;welcome&quot;</strong>
+            </>
+          ),
```
migueloller, line 16: This is a weird prop interface. Why not have `banner` itself be a `ReactNode`?

apps/web/vibes/soul/primitives/navigation/index.tsx
```diff
@@ -507,19 +243,361 @@ export const Navigation = forwardRef(function Navigation(
...
+  }, [lastResult])
+
+  useEffect(() => {
+    if (debouncedQuery === '') return
+
+    startTransition(() => {
+      const formData = new FormData()
+      formData.append(searchParamName, debouncedQuery)
+
+      formAction(formData)
```
migueloller, line 419: I still don't love that we're using actions for search but I don't have a better alternative yet...

migueloller (approved):
LGTM 👍🏻 

I really do think that we'll figure something better longterm but for now we'll have to live with this...

PR #296, opened 2024-11-07
fix(Navigation): make `localeAction` optional, similar to the list of locales

agurtovoy:


apps/web/vibes/soul/primitives/navigation/index.tsx
```diff
@@ -78,7 +78,7 @@ type Props = {
   links: Link[]
   locales?: Locale[]
   activeLocaleId?: string
-  localeAction: LocaleAction
+  localeAction?: LocaleAction
```
agurtovoy, line 81: We can get fancy here and require that either both `locales` and `localeAction` are there or not, but I didn't want to mess with it right now.

migueloller (approved):
LGTM 👍🏻

PR #318, opened 2024-11-08
fix: cursor navigation by remove unused param

jorgemoya:
## What/Why?
For cursor navigation we either want an `after` or `before`, not both.

Before:

https://github.com/user-attachments/assets/e7189cd1-ebbf-47df-8197-211073c94ed5

After:

https://github.com/user-attachments/assets/1a76128f-846a-418e-913d-368e724c8254


apledger (commented):
LGTM!

PR #341, opened 2024-11-20
feat: add streamable props in Reviews section

jorgemoya:
- Wraps `<Reviews />` in `Suspense` boundary and internally resolves Promises if passed.
- Streams in data inside components for total count and average rating.
- Adds missing `reviewsLabel`.

### DIFF
--- a/apps/web/vibes/soul/sections/reviews/index.tsx
+++ b/apps/web/vibes/soul/sections/reviews/index.tsx
@@ -1,41 +1,76 @@
+import { Suspense } from 'react'
+
+import { Streamable } from '@/vibes/soul/lib/streamable'
+import { mapStreamable } from '@/vibes/soul/lib/streamable/server'
 import { CursorPagination, CursorPaginationInfo } from '@/vibes/soul/primitives/cursor-pagination'
 import { Rating } from '@/vibes/soul/primitives/rating'
 import { StickySidebarLayout } from '@/vibes/soul/sections/sticky-sidebar-layout'
 
-interface Props {
-  reviews: {
-    id: string
-    rating: number
-    review: string
-    name: string
-    date: string
-  }[]
-  averageRating: number
-  totalCount?: number
-  paginationInfo?: CursorPaginationInfo
-  emptyStateMessage?: string
+interface Review {
+  id: string
+  rating: number
+  review: string
+  name: string
+  date: string
 }
 
-export function Reviews({
-  reviews,
-  averageRating,
-  totalCount,
-  paginationInfo,
+interface Props {
+  reviews: Streamable<Review[]>
+  averageRating: Streamable<number>
+  totalCount?: Streamable<number>
+  paginationInfo?: Streamable<CursorPaginationInfo>
+  emptyStateMessage?: string
+  reviewsLabel?: string
+}
+
+async function ReviewsImpl({
+  reviews: streamableReviews,
+  averageRating: streamableAverageRating,
+  totalCount: streamableTotalCount,
+  paginationInfo: streamablePaginationInfo,
   emptyStateMessage,
+  reviewsLabel = 'Reviews',
 }: Readonly<Props>) {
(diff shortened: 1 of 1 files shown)


apps/web/vibes/soul/sections/reviews/index.tsx
```diff
@@ -1,41 +1,76 @@
...
   emptyStateMessage?: string
+  reviewsLabel?: string
 }
 
-export function Reviews({
-  reviews,
-  averageRating,
-  totalCount,
-  paginationInfo,
+async function ReviewsImpl({
```
jorgemoya, line 26: Not sure what is the naming convention for internal components.

apps/web/vibes/soul/sections/reviews/index.tsx
```diff
@@ -1,41 +1,76 @@
...
+  reviews: streamableReviews,
+  averageRating: streamableAverageRating,
+  totalCount: streamableTotalCount,
+  paginationInfo: streamablePaginationInfo,
   emptyStateMessage,
+  reviewsLabel = 'Reviews',
 }: Readonly<Props>) {
+  const reviews = await mapStreamable(streamableReviews, resolvedReviews => resolvedReviews)
```
jorgemoya, line 34: I tried to wrap this in a `<Suspense />`, but below we use this value to know if we should render the empty state, so unfortunately I need to resolve here. Any ideas?

apps/web/vibes/soul/sections/reviews/index.tsx
```diff
@@ -110,3 +153,11 @@ export function ReviewsSkeleton() {
...
   )
 }
+
+export function Reviews({ reviewsLabel = 'Reviews', ...props }: Props) {
+  return (
+    <Suspense fallback={<ReviewsSkeleton />}>
+      <ReviewsImpl {...props} reviewsLabel={reviewsLabel} />
+    </Suspense>
+  )
+}
```
jorgemoya, line 163: This is only needed because I resolve reviews to know if I want to render empty component or not. See above.

apps/web/vibes/soul/sections/reviews/index.tsx
```diff
@@ -85,12 +128,12 @@ export function ReviewsEmptyState({
   )
 }
 
-export function ReviewsSkeleton() {
+export function ReviewsSkeleton({ reviewsLabel = 'Reviews' }: { reviewsLabel?: string }) {
```
jorgemoya, line 131: I think exporting the Skeleton is still a great idea for those use cases when users want to stream in the entire data.

apps/web/vibes/soul/sections/reviews/index.tsx
migueloller: If we're using `await` there's no need to use `mapStreamable` since we can just do this:

```ts
const reviews = await streamableReviews
```

That being said, using `mapStreamable` should remove the need to have a `ReviewsImpl` and another `Review` component that wraps it

migueloller: You could do something like this:

```ts
return (
  <Suspense fallback={<Skeleton />}>
    {mapStreamable(streamableReview, reviews => reviews.length === 0 ? <EmptyState /> : <ReviewsList reviews={reviews} />)}
  </Suspense>
)
```

What you have right now works completely fine, though, and I'm not too worried about the VIBES code being perfect as much as I am about the pattern itself being close to what we need to support

migueloller (approved):
Looks good 👍🏻 

By the way, looks like you're still opening this PR from a fork instead of a branch from the repo, which is resulting in Vercel deployments not running automatically and asking for authorization. I wanted to play around with the preview but wasn't able to. Do you mind changing that?

apps/web/vibes/soul/sections/reviews/index.tsx
jorgemoya: I get it now 👍 . Let me push a quick refactor.

jorgemoya:
Will move away from using a `fork` after this PR.

PR #346, opened 2024-11-22
feat: ENG-6931 export FeaturedProductsCarouselSkeleton

jorgemoya:
Exports a Skeleton.

Still missing empty state.

Screenshot 2024-11-22 at 12 42 33 PM

### DIFF
--- a/apps/web/vibes/soul/primitives/products-carousel/index.tsx
+++ b/apps/web/vibes/soul/primitives/products-carousel/index.tsx
@@ -29,7 +29,7 @@ export function ProductsCarousel({
       <CarouselContent className="mb-10">
         <Stream
           value={streamableProducts}
-          fallback={<ProductsCarouselSkeleton className={className} message={emptyStateMessage} />}
+          fallback={<ProductsCarouselSkeleton className={className} />}
         >
           {products => {
             if (products.length === 0) {
--- a/apps/web/vibes/soul/primitives/products-carousel/index.tsx
+++ b/apps/web/vibes/soul/primitives/products-carousel/index.tsx
@@ -27,13 +27,10 @@ export function ProductsCarousel({
   return (
     <Carousel className={className}>
       <CarouselContent className="mb-10">
-        <Stream
-          value={streamableProducts}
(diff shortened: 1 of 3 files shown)


apps/web/vibes/soul/primitives/products-carousel/index.tsx
```diff
@@ -27,13 +27,10 @@ export function ProductsCarousel({
   return (
     <Carousel className={className}>
       <CarouselContent className="mb-10">
-        <Stream
-          value={streamableProducts}
-          fallback={<ProductsCarouselSkeleton className={className} message={emptyStateMessage} />}
-        >
+        <Stream value={streamableProducts} fallback={<ProductsCarouselSkeleton />}>
```
migueloller, line 30: I think the reason we were passing the class was to control with in Makeswift. Might need to rethink later by wrapping in div and passing class there

apps/web/vibes/soul/sections/featured-products-carousel/index.tsx
```diff
@@ -45,3 +45,31 @@ export function FeaturedProductsCarousel({ title, description, cta, products }:
     </section>
   )
 }
+
+export function FeaturedProductsCarouselSkeleton({
+  title,
+  description,
+  cta,
```
migueloller, line 52: Are these going to always be available? i.e., not streamable?

migueloller (approved):

apps/web/vibes/soul/sections/featured-products-carousel/index.tsx
jorgemoya: In Catalyst, yes, in general, maybe not?

apps/web/vibes/soul/primitives/products-carousel/index.tsx
jorgemoya: It is weird that the `className` is being passed to the `Carousel` component but that same to the Skeleton (which is nested within `CarouselContent`). Seems like an oversight to me?

migueloller: That could be a mistake

PR #354, opened 2024-11-25
fix: slide not in view when element inside is focused

hunterbecton:
## What/Why?

- Fixes issue where the slide was not visible when the button was focused
- Replaces `text-pretty` with `text-balance` for text-wrapping

## Testing

https://github.com/user-attachments/assets/96060633-40b8-4585-8572-5d4d0019c6f1

### DIFF
--- a/apps/web/vibes/soul/sections/slideshow/index.tsx
+++ b/apps/web/vibes/soul/sections/slideshow/index.tsx
@@ -113,16 +113,14 @@ export function Slideshow({ slides, interval = 5000, className }: Props) {
   }, [emblaApi, playCount])
 
   return (
-    <section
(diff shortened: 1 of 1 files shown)


migueloller (changes requested):
Looks like there's changes in `pnpm-lock.yaml` that shouldn't be there?

Also, can we `git rebase` and squash the two commits? Can also do this at merge time in GitHub using "Squash and merge".

migueloller:
Also, it looks like we didn't pull in the `text-balance` change

hunterbecton (commented):
Added the `text-balance` Tailwind class.

migueloller (approved):

PR #355, opened 2024-12-03
Fix Sitemap and Algolia Search

jamesqquick:
- update sitemap to ignore paths with `/preview`
- remove write API key for algolia and use read only key
- ignore `robots.txt` so it is generated on build
- update package.json `build` script to also call `sitemap` script (wasn't be called automatically in `postbuild` script)
- reference `VERCEL_URL` environment variable instead of hard-coded vibes.site url - this will allow us to test crawlers on preview deploy links if we want to

### DIFF
--- a/apps/web/components/navigation/header.tsx
+++ b/apps/web/components/navigation/header.tsx
@@ -6,6 +6,7 @@ import Image from 'next/image'
 import '@docsearch/css'
 
 import { ModeToggle } from '@/components/ui/mode-toggle'
+import { env } from '@/lib/env'
 
 import { ChapterSelect } from './chapter-select'
 import { GroupLink } from './group-link'
@@ -65,9 +66,9 @@ export function Header({ chapters, chapterSlug }: Props) {
           </Button> */}
 
           <DocSearch
-            appId="Q9L04M9AMF"
-            indexName="vibes"
-            apiKey="931574c95fe03313ea73bd5a77286fee"
+            appId={env.NEXT_PUBLIC_ALGOLIA_APP_ID}
+            indexName={env.NEXT_PUBLIC_ALGOLIA_INDEX_NAME}
+            apiKey={env.NEXT_PUBLIC_ALGOLIA_SEARCH_API_KEY}
           />
 
           <ModeToggle />
--- a/apps/web/components/navigation/header.tsx
+++ b/apps/web/components/navigation/header.tsx
@@ -67,7 +67,7 @@ export function Header({ chapters, chapterSlug }: Props) {
           <DocSearch
             appId="Q9L04M9AMF"
             indexName="vibes"
-            apiKey="ef510190238e6bf2ae942046fd83e4b9"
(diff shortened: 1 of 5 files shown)


apps/web/package.json
```diff
@@ -5,11 +5,11 @@
   "license": "MIT",
   "scripts": {
     "dev": "next dev --turbo",
-    "build": "next build",
+    "build": "next build && pnpm sitemap",
```
jamesqquick, line 8: Originally, it didn't appear that the `postbuild` script was being run automatically. I tried converting it to just echo something out, and it never showed up. I also noticed that sitemap files weren't being generated in my preview deploy. 

Is there a better way to ensure that the `sitemap` command is being run?

migueloller: For pnpm we'd have to [enable this functionality](https://pnpm.io/cli/run#enable-pre-post-scripts).

We could add an `.npmrc` with the following contents:

```ini
enable-pre-post-scripts=true
```

I think that having it like you have it in the `build` script explicitly is better, though!

apps/web/components/navigation/header.tsx
```diff
@@ -67,7 +67,7 @@ export function Header({ chapters, chapterSlug }: Props) {
           <DocSearch
             appId="Q9L04M9AMF"
             indexName="vibes"
-            apiKey="ef510190238e6bf2ae942046fd83e4b9"
+            apiKey="931574c95fe03313ea73bd5a77286fee"
```
migueloller, line 70: Is there a reason we're not using an environment variable for this? It feels like this should be a configuration value, even if it's `NEXT_PUBLIC_`

migueloller: Same goes for the `appId`

apps/web/package.json
```diff
@@ -5,11 +5,11 @@
...
     "start": "next start",
     "lint": "next lint",
     "icons": "npx @svgr/cli --no-runtime-config --typescript --replace-attr-values \"#64de21=currentColor\" --out-dir icons/generated -- icons/svg && npx prettier icons/generated --write",
-    "postbuild": "next-sitemap",
+    "sitemap": "next-sitemap",
```
migueloller, line 12: We probably don't need this script and could update the `build` script to be `next build && next-sitemap`

migueloller (dismissed):
Looks good. I think it would be nice if we moved the DocSearch configuration to environment variables, our choice for configuration at the moment. Also, we might want to consider using a different solution for sitemaps given `next-sitemap` hasn't been maintained for over a year now and that Next.js now has first-class supports for sitemaps via `sitemap.{xml,js,ts}` and `generateSitemaps`

apps/web/components/navigation/header.tsx
jamesqquick: Yep, we should! Updating now.

apps/web/package.json
jamesqquick: Ah gotcha. Let me try that!

migueloller (approved):
Nice!

PR #356, opened 2024-12-03
feat: add streamable props to Header

jorgemoya:
+ Add streaming to the Header props:
  + logo
  + cartCount
  + links
+ Fix image props
+ Add missing aria label props

I don't think we need to stream in locales because this should be hardcoded afaik.

Update examples to showcase this behavior.

https://github.com/user-attachments/assets/67b22667-75c8-45ea-b890-626ab51478c5

### DIFF
--- a/apps/web/vibes/soul/examples/sections/header-section/electric.tsx
+++ b/apps/web/vibes/soul/examples/sections/header-section/electric.tsx
@@ -8,7 +8,7 @@ export default function Preview() {
     setTimeout(() => res(navigationLinks), 3000)
   )
   const cartCountPromise = new Promise<number>(res => setTimeout(() => res(3), 2000))
-  const logoPromise = new Promise<string | { src?: string; alt: string }>(res =>
+  const logoPromise = new Promise<string | { src: string; alt: string }>(res =>
     setTimeout(() => res(logo), 1000)
   )
 
--- a/apps/web/vibes/soul/examples/sections/header-section/electric.tsx
+++ b/apps/web/vibes/soul/examples/sections/header-section/electric.tsx
@@ -4,6 +4,14 @@ import { logo, navigationLinks } from '@/vibes/soul/examples/primitives/navigati
 import { HeaderSection } from '@/vibes/soul/sections/header-section'
 
 export default function Preview() {
+  const linksPromise = new Promise<{ href: string; label: string }[]>(res =>
+    setTimeout(() => res(navigationLinks), 3000)
+  )
+  const cartCountPromise = new Promise<number>(res => setTimeout(() => res(3), 2000))
+  const logoPromise = new Promise<string | { src?: string; alt: string }>(res =>
+    setTimeout(() => res(logo), 1000)
+  )
+
   return (
     <>
       <HeaderSection
@@ -16,8 +24,9 @@ export default function Preview() {
           ),
         }}
         navigation={{
-          links: navigationLinks,
-          logo,
+          links: linksPromise,
+          logo: logoPromise,
+          cartCount: cartCountPromise,
           cartHref: '#',
           accountHref: '#',
           activeLocaleId: 'US',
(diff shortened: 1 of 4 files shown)


apps/web/vibes/soul/examples/sections/header-section/electric.tsx
```diff
@@ -4,6 +4,14 @@ import { logo, navigationLinks } from '@/vibes/soul/examples/primitives/navigati
 import { HeaderSection } from '@/vibes/soul/sections/header-section'
 
 export default function Preview() {
+  const linksPromise = new Promise<{ href: string; label: string }[]>(res =>
```
jamesqquick, line 7: I may not have full context but why are we adding hard coded delays here? Obviously, we have streaming if loading the data takes some amount of time, but would we manually add the delay?

apps/web/vibes/soul/primitives/navigation/index.tsx
```diff
@@ -79,13 +80,13 @@ interface Props<S extends SearchResult> {
   className?: string
   isFloating?: boolean
   accountHref: string
-  cartCount?: number
+  cartCount?: Streamable<number>
```
migueloller, line 83: I think this should be `Streamable<number | null>` since the prop is optional

apps/web/vibes/soul/primitives/navigation/index.tsx
```diff
@@ -79,13 +80,13 @@ interface Props<S extends SearchResult> {
...
   cartHref: string
-  links: Link[]
+  links: Streamable<Link[]>
   locales?: Locale[]
   activeLocaleId?: string
   localeAction?: LocaleAction
-  logo?: string | { src?: string; alt: string }
+  logo?: Streamable<string | { src: string; alt: string }>
```
migueloller, line 89: We should add `| null` here since the prop is optional

apps/web/vibes/soul/primitives/navigation/index.tsx
```diff
@@ -373,26 +431,44 @@ export const Navigation = forwardRef(function Navigation<S extends SearchResult>
...
               </Popover.Portal>
             </Popover.Root>
+          ) : (
+            <Link
+              aria-label={searchLabel}
+              className="rounded-lg p-1.5 ring-primary transition-colors focus-visible:outline-0 focus-visible:ring-2 @4xl:hover:bg-contrast-100"
+              href={searchHref}
+            >
+              <Search size={20} strokeWidth={1} />
+            </Link>
```
migueloller, line 441: So what we're doing here is if the action isn't provided we fall back to the href when clicking the search icon? I like it 👍🏻

migueloller (changes requested):

apps/web/vibes/soul/primitives/navigation/index.tsx
jorgemoya: Can you expand? If it's optional that means we just don't pass anything, no?

apps/web/vibes/soul/primitives/navigation/index.tsx
jorgemoya: I noticed `searchHref` was required but `searchAction` was not, but if `searchAction` was undefined we hide the icon completely. One solution was just to make `searchHref` optional. I like this solution better.

apps/web/vibes/soul/primitives/navigation/index.tsx
migueloller: Sometimes you don't know until the promise resolves but then the types don't let the promise have a null in it

apps/web/vibes/soul/primitives/navigation/index.tsx
migueloller: Yeah, that's good. I think later we can actually make `searchAction` be `Action | string` like Next.js 15 Form API does. I think that would be cleanest

migueloller (approved):

apps/web/vibes/soul/primitives/navigation/index.tsx
jorgemoya: True!

PR #359, opened 2024-12-04
fix: require id for banner component

jorgemoya:
- Require an id that is used to hide or show banner.
- Update examples.
- Compute the height of the banner instead of using state.


migueloller (approved):
The change to require an ID looks good to me. The way we handle height I think was not great and it still isn't that great.

migueloller (approved):

PR #360, opened 2024-12-04
feat: add Toaster primitive for Soul

jorgemoya:
- Adds the `Toaster` primitive that uses [Sonner](apps/web/vibes/soul/sections/subscribe/index.tsx) for toasts.
- Customize Toaster.
- Updated the docs to include this new primitive.
- Add primitive to preview to enable use of `toast` in other components.

### Demo
https://github.com/user-attachments/assets/635b1c8d-9862-4a69-b9f6-bd9905db4d34

`Sonner` has it on the corner since the toast can have an action on the right side of the toast. 

Screenshot 2024-12-04 at 4 38 09 PM

The proposed designs has the close/dismiss button as an action:

image

I feel it may be important to retain the action prop api for Toast so I think it is a better implementation to keep the close button in the corner as it currently works from the box in Sonner, but I want to hear what @andrewreifman might have to say on this.

Another option is to have a custom `toast` function that does not accept actions.

### DIFF
--- a/apps/web/app/(preview)/layout.tsx
+++ b/apps/web/app/(preview)/layout.tsx
@@ -1,5 +1,7 @@
-import { ZoomListener } from '../zoom-listener';
 import './style/globals.css';
+import { Toaster } from '@/vibes/soul/primitives/toaster';
+
+import { ZoomListener } from '../zoom-listener';
 
 export const metadata = {
   title: 'Vibes preview',
@@ -10,7 +12,11 @@ export default function RootLayout({ children }: { children: React.ReactNode }) 
   return (
     <html lang="en">
       <ZoomListener />
-      <body>{children}</body>
+
+      <body>
+        <Toaster closeButton={true} position="top-right" />
+        {children}
+      </body>
     </html>
   );
 }
--- a/apps/web/app/(preview)/layout.tsx
+++ b/apps/web/app/(preview)/layout.tsx
@@ -14,6 +14,9 @@ export default function RootLayout({ children }: { children: React.ReactNode }) 
       <ZoomListener />
 
       <body>
+        {/* TODO: This is currently hardcoded to the Soul vibe. 
+        We want to make it so that each vibe has its own preview 
+        layout. */}
         <Toaster position="top-right" />
         {children}
       </body>
--- a/apps/web/app/(preview)/layout.tsx
+++ b/apps/web/app/(preview)/layout.tsx
@@ -14,7 +14,7 @@ export default function RootLayout({ children }: { children: React.ReactNode }) 
       <ZoomListener />
 
       <body>
-        <Toaster closeButton={true} position="top-right" />
+        <Toaster position="top-right" />
         {children}
       </body>
     </html>
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -80,6 +80,7 @@
     "remark-mdx-frontmatter": "^4.0.0",
     "resend": "^3.2.0",
     "server-only": "^0.0.1",
+    "sonner": "^1.7.0",
     "tailwind-merge": "^2.2.1",
     "utils": "link:@/lib/utils",
     "zod": "^3.23.8"
--- a/apps/web/vibes/soul/docs/toaster.mdx
+++ b/apps/web/vibes/soul/docs/toaster.mdx
@@ -1,0 +1,5 @@
+---
+title: Toaster 
+preview: toaster-example
+previewSize: sm
+---
--- a/apps/web/vibes/soul/examples.ts
+++ b/apps/web/vibes/soul/examples.ts
@@ -840,4 +840,11 @@ export const examples = [
     files: ['examples/primitives/textarea/index.tsx'],
     component: lazy(() => import('./examples/primitives/textarea')),
   },
+  {
+    name: 'toaster-example',
+    dependencies: [],
+    registryDependencies: ['toaster'],
+    files: ['examples/primitives/toaster/index.tsx'],
+    component: lazy(() => import('./examples/primitives/toaster')),
+  },
 ] satisfies Components;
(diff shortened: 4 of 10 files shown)


apps/web/app/(preview)/layout.tsx
```diff
@@ -10,7 +12,11 @@ export default function RootLayout({ children }: { children: React.ReactNode })
   return (
     <html lang="en">
       <ZoomListener />
-      <body>{children}</body>
+
+      <body>
+        <Toaster position="top-right" closeButton={true} />
+        {children}
+      </body>
```
jorgemoya, line 19: This is not final. What would be the best way to implement the VIBES Toaster for all examples for a specific vibe? Or do we need to add this on a per example basis? Open to suggestions.

andrewreifman:
@jorgemoya Don't love some of the styling at the moment. I would prefer to place an optional action button next to the close icon instead of using Sonner's corner version. In Figma, I created a ghost variant that we never applied to the codebase, as well as an "x-small" size option. I updated the [Figma mockup](https://www.figma.com/design/GAKpWYdZ0ZVOSFtNPHliJM/Soul-Vibe?node-id=2786-48748&t=qca2YvH6oEucMQoI-1) to include this, as well as removing the left icon from all variants. This was something Monogram had in there initially that I kept, but I'm wondering if we should just remove the left icon to simplify it.

Let me know what you think.

jorgemoya:
@andrewreifman This works for me! I'll be updating the designs to match this.

jorgemoya:
Updated designs:

Screenshot 2024-12-04 at 5 52 47 PM

Screenshot 2024-12-04 at 6 06 37 PM

Screenshot 2024-12-04 at 6 10 35 PM

migueloller:
Was playing around with it and I think it looks pretty good. There doesn't seem to be any active state on the close buttons, though. @andrewreifman, can we get those states mocked up? The current mockups seem quite lacking in terms of interaction states

apps/web/app/(preview)/layout.tsx
migueloller: We'll probably want to have a preview root layout for each VIBE. Maybe we can accomplish that using [parallel routes](https://nextjs.org/docs/app/building-your-application/routing/parallel-routes)?

I'm ok cutting this corner for now but let's add a comment in this file to make it clear that this is hardcoded to the Soul VIBE's Toaster primitive and we would want to have each VIBE define its own layout.

Another idea is to export the layout from within each VIBE and then build a map of `lazy` imports and then use the VIBE slug to get the appropriate layout. We'd have to render it further down the routes so we can access the parameter, though.

migueloller (dismissed):
LGTM 👍🏻

andrewreifman:
> Was playing around with it and I think it looks pretty good. There doesn't seem to be any active state on the close buttons, though. @andrewreifman, can we get those states mocked up? The current mockups seem quite lacking in terms of interaction states

CleanShot 2024-12-05 at 11 36 47@2x

Ok, did one more round of design revisions. Here are some of the things I updated:
- Modified some general styles on the alert like width and padding
- Added a subtle border and drop shadow
- Added a white variant that I think we could use as the default
- Added subtext under the main message
- Converted the optional button and close to use actual button component. For the button component, there is a ghost variant, as well as some size options that are not available in the current state of the code in VIBES. That will have to be updated as well. These buttons will have the states included.

@migueloller @jorgemoya Let me know if there is anything else missing in the mockups.

migueloller:
> > Was playing around with it and I think it looks pretty good. There doesn't seem to be any active state on the close buttons, though. @andrewreifman, can we get those states mocked up? The current mockups seem quite lacking in terms of interaction states
> 
> CleanShot 2024-12-05 at 11 36 47@2x
> 
> Ok, did one more round of design revisions. Here are some of the things I updated:
> 
> * Modified some general styles on the alert like width and padding
> * Added a subtle border and drop shadow
> * Added a white variant that I think we could use as the default
> * Added subtext under the main message
> * Converted the optional button and close to use actual button component. For the button component, there is a ghost variant, as well as some size options that are not available in the current state of the code in VIBES. That will have to be updated as well. These buttons will have the states included.
> 
> @migueloller @jorgemoya Let me know if there is anything else missing in the mockups.

@andrewreifman, this looks great! I think what we're missing are the active and focus-visible states. The mockups seem to only include the hover states.

andrewreifman:
> @andrewreifman, this looks great! I think what we're missing are the active and focus-visible states. The mockups seem to only include the hover states.

@migueloller We should be using the `Button` component for those actions. Do you feel like it's necessary to show that in this mockup when the Button Figma already includes those states? And do we need focus-visible states for the toasts themselves?

Also, hot take, but I don't like including active states for most components. I think it's unnecessary to include unless there's an accessibility reason I'm unaware of.

jorgemoya:
Hey @migueloller, I updated how this component is used to allow the use of our custom Alert primitives. Let me know what you think, I'm always open to reverting the latest change.

migueloller:
> > @andrewreifman, this looks great! I think what we're missing are the active and focus-visible states. The mockups seem to only include the hover states.
> 
> @migueloller We should be using the `Button` component for those actions. Do you feel like it's necessary to show that in this mockup when the Button Figma already includes those states? And do we need focus-visible states for the toasts themselves?
> 
> Also, hot take, but I don't like including active states for most components. I think it's unnecessary to include unless there's an accessibility reason I'm unaware of.

If we're reusing the Button component, then it's not necessary, no.

And I won't argue with you if you think active states aren't useful. I personally disagree as I think that UI should provide feedback on interaction but I'll delegate to you here

migueloller (commented):
Love it!

jorgemoya:
Will use vibes button component once this is merged https://github.com/makeswift/vibes/pull/364

migueloller (approved):

PR #364, opened 2024-12-05
Andrew/button options

andrewreifman:
I added a new 'ghost' variant to the Button and ButtonLink components. I also added 2 more size options for each component: "x-small" and "icon-small".

CleanShot 2024-12-05 at 15 43 12@2x

### DIFF
--- a/apps/web/vibes/soul/examples/primitives/button-link/index.tsx
+++ b/apps/web/vibes/soul/examples/primitives/button-link/index.tsx
@@ -2,17 +2,21 @@ import { ButtonLink } from '@/vibes/soul/primitives/button-link';
 
 export default function Preview() {
   return (
-    <div className="flex h-screen flex-col items-center justify-center gap-y-6">
+    <div className="flex h-screen flex-col items-center justify-center gap-y-5">
       <ButtonLink href="#" variant="primary">
-        Shop Now
+        Shop now
       </ButtonLink>
 
       <ButtonLink href="#" variant="secondary">
-        Shop Now
+        Shop now
       </ButtonLink>
 
       <ButtonLink href="#" variant="tertiary">
-        Shop Now
+        Shop now
+      </ButtonLink>
+
+      <ButtonLink href="#" variant="ghost">
+        Shop now
       </ButtonLink>
     </div>
   );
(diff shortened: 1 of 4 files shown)


drewpledger (approved):
LGTM

migueloller (dismissed):
@andrewreifman, this LGTM. I will say, though, I think our buttons need active states (i.e., what they look like while being pressed). Right now it feels weird they don't respond to clicks.

Also, can we add the size variants to the example page in the docs? I was testing things out in the preview deployment but the example doesn't show the new variants so I wasn't able to test those.

jorgemoya (approved):

andrewreifman:
> @andrewreifman, this LGTM. I will say, though, I think our buttons need active states (i.e., what they look like while being pressed). Right now it feels weird they don't respond to clicks.
> 
> Also, can we add the size variants to the example page in the docs? I was testing things out in the preview deployment but the example doesn't show the new variants so I wasn't able to test those.

I think I've never liked active states because they feel like this glitchy thing you see for 50ms. I feel like me actually clicking a button, plus the action that happens right after is enough affordance to let me know I clicked it. And if there is an extended delay on an action, we have the loading state on the Button.

We've talked about adding more previews within each component's doc, so we could display other props like size, but I could go ahead and add it to the main preview for now.

migueloller:
> > @andrewreifman, this LGTM. I will say, though, I think our buttons need active states (i.e., what they look like while being pressed). Right now it feels weird they don't respond to clicks.
> > Also, can we add the size variants to the example page in the docs? I was testing things out in the preview deployment but the example doesn't show the new variants so I wasn't able to test those.
> 
> I think I've never liked active states because they feel like this glitchy thing you see for 50ms. I feel like me actually clicking a button, plus the action that happens right after is enough affordance to let me know I clicked it. And if there is an extended delay on an action, we have the loading state on the Button.
> 
> We've talked about adding more previews within each component's doc, so we could display other props like size, but I could go ahead and add it to the main preview for now.

The active state is present while the mouse is pressed down, though. Maybe you're a super fast clicker, but not everybody is 😄

jorgemoya:
+1 on updating the examples to play with variants

andrewreifman:
@migueloller For you my friend. It's the same for the Button preview.

CleanShot 2024-12-05 at 16 29 32@2x

migueloller (dismissed):
Nice work @andrewreifman ❤️

.github/workflows/dismiss_stale_approvals.yml
migueloller, line 1: This file shouldn't be in this PR since it's already in `main` 🤔

migueloller (dismissed):

migueloller (approved):

jorgemoya (approved):

PR #366, opened 2024-12-05
feat: update subscribe section to show toast and intl props

jorgemoya:
+ Update component to show toast on success
+ Add `successMessage` prop
+ Add `dismissLabel` prop
+ Change revalidation `onBlur` (this felt better for me)

As a side note: 1password is trying to save this field.

https://github.com/user-attachments/assets/c65ca487-d4fd-424d-b351-3e0bd76fcb19

### DIFF
--- a/apps/web/vibes/soul/examples/primitives/inline-email-form/actions.ts
+++ b/apps/web/vibes/soul/examples/primitives/inline-email-form/actions.ts
@@ -3,17 +3,20 @@ import { parseWithZod } from '@conform-to/zod';
 
 import { schema } from '@/vibes/soul/primitives/inline-email-form/schema';
 
-export async function action(lastResult: SubmissionResult | null, formData: FormData) {
+export async function action(
+  _lastResult: { lastResult: SubmissionResult | null; },
+  formData: FormData,
+) {
   'use server';
 
   const submission = parseWithZod(formData, { schema });
 
   if (submission.status !== 'success') {
-    return submission.reply({ formErrors: ['Boom!'] });
+    return { lastResult: submission.reply({ formErrors: ['Boom!'] }) };
   }
 
   // Simulate a network request
   await new Promise((resolve) => setTimeout(resolve, 1000));
 
-  return { ...submission.reply(), successMessage: 'Subscribed!' };
+  return { lastResult: submission.reply(), successMessage: 'Subscribed!' };
 }
--- a/apps/web/vibes/soul/examples/primitives/inline-email-form/actions.ts
+++ b/apps/web/vibes/soul/examples/primitives/inline-email-form/actions.ts
@@ -15,7 +15,5 @@ export async function action(lastResult: SubmissionResult | null, formData: Form
   // Simulate a network request
   await new Promise((resolve) => setTimeout(resolve, 1000));
 
-  // const user = await logIn(submission.value)
-
-  return submission.reply({ resetForm: true });
+  return submission.reply();
 }
--- a/apps/web/vibes/soul/examples/primitives/inline-email-form/actions.ts
+++ b/apps/web/vibes/soul/examples/primitives/inline-email-form/actions.ts
@@ -15,5 +15,5 @@ export async function action(lastResult: SubmissionResult | null, formData: Form
   // Simulate a network request
   await new Promise((resolve) => setTimeout(resolve, 1000));
 
-  return {...submission.reply({ resetForm: true }), successMessage: 'Subscribed!'};
+  return { ...submission.reply({ resetForm: true }), successMessage: 'Subscribed!' };
 }
--- a/apps/web/vibes/soul/examples/primitives/inline-email-form/actions.ts
+++ b/apps/web/vibes/soul/examples/primitives/inline-email-form/actions.ts
@@ -4,7 +4,7 @@ import { parseWithZod } from '@conform-to/zod';
 import { schema } from '@/vibes/soul/primitives/inline-email-form/schema';
 
 export async function action(
-  _lastResult: { lastResult: SubmissionResult | null; },
+  _lastResult: { lastResult: SubmissionResult | null },
   formData: FormData,
 ) {
   'use server';
--- a/apps/web/vibes/soul/examples/primitives/inline-email-form/actions.ts
+++ b/apps/web/vibes/soul/examples/primitives/inline-email-form/actions.ts
@@ -15,5 +15,5 @@ export async function action(lastResult: SubmissionResult | null, formData: Form
   // Simulate a network request
   await new Promise((resolve) => setTimeout(resolve, 1000));
 
-  return submission.reply();
+  return {...submission.reply({ resetForm: true }), successMessage: 'Subscribed!'};
 }
--- a/apps/web/vibes/soul/examples/primitives/inline-email-form/actions.ts
+++ b/apps/web/vibes/soul/examples/primitives/inline-email-form/actions.ts
@@ -15,5 +15,5 @@ export async function action(lastResult: SubmissionResult | null, formData: Form
   // Simulate a network request
   await new Promise((resolve) => setTimeout(resolve, 1000));
 
-  return { ...submission.reply({ resetForm: true }), successMessage: 'Subscribed!' };
+  return { ...submission.reply(), successMessage: 'Subscribed!' };
 }
--- a/apps/web/vibes/soul/primitives/inline-email-form/index.tsx
+++ b/apps/web/vibes/soul/primitives/inline-email-form/index.tsx
@@ -25,9 +25,13 @@ export function InlineEmailForm({
   className?: string;
   placeholder?: string;
   submitLabel?: string;
-  action: Action<(SubmissionResult & { successMessage?: string }) | null, FormData>;
+  action: Action<{ lastResult: SubmissionResult | null; successMessage?: string }, FormData>;
 }) {
-  const [lastResult, formAction, isPending] = useActionState(action, null);
+  const [{ lastResult, successMessage }, formAction, isPending] = useActionState(action, {
+    lastResult: null,
+  });
+
+  console.log('lastResult', lastResult);
 
   const [form, fields] = useForm({
     lastResult,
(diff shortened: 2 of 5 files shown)


apps/web/vibes/soul/primitives/inline-email-form/index.tsx
```diff
@@ -21,10 +22,14 @@ export function InlineEmailForm({
   action,
   submitLabel = 'Submit',
   placeholder = 'Enter your email',
+  successMessage,
+  dismissLabel,
 }: {
   className?: string;
   placeholder?: string;
   submitLabel?: string;
+  successMessage: string;
```
jorgemoya, line 31: Is this the best way to pass in a `success` message for the toast? Open to suggestions.

Wondering if this is also that can be leveraged from conform?

migueloller: Yes, we should be using `submission.reply` from the server and have the payload return a message on success

migueloller:
> As a side note: 1password is trying to save this field.

We can add `data-1p-ignore` to disable it

apps/web/vibes/soul/examples/primitives/inline-email-form/actions.ts
```diff
@@ -15,7 +15,5 @@ export async function action(lastResult: SubmissionResult | null, formData: Form
   // Simulate a network request
   await new Promise((resolve) => setTimeout(resolve, 1000));
 
-  // const user = await logIn(submission.value)
-
-  return submission.reply({ resetForm: true });
+  return submission.reply();
```
migueloller, line 18: We should still use `resetForm: true` here, no? We wouldn't want the user to subscribe twice accidentally?

apps/web/vibes/soul/primitives/inline-email-form/index.tsx
```diff
@@ -35,15 +40,15 @@ export function InlineEmailForm({
...
 
   useEffect(() => {
-    if (lastResult?.error) {
-      console.log(lastResult.error);
+    if (lastResult?.status === 'success') {
+      toast.success(successMessage, { dismissLabel });
       return;
     }
-  }, [lastResult]);
+  }, [dismissLabel, lastResult, successMessage]);
```
migueloller, line 51: I feel like when submission succeeds we should have the success state be inline on the form, like the error instead of using a toast. cc: @andrewreifman. I like what we do in https://vibes.site.

migueloller (approved):
Looking good.

I think the contrast on the error message is not great, though. That red on dark green is barely visible. Do we need to have a lighter red for better contrast or perhaps field errors should have a well @andrewreifman?

apps/web/vibes/soul/primitives/inline-email-form/index.tsx
migueloller: Sorry, not `submission.reply`, but the action payload like we have done in other form actions

migueloller:
I don't particularly like how the field shifts as well when there's an error. Seems like this needs some design love @andrewreifman

apps/web/vibes/soul/examples/primitives/inline-email-form/actions.ts
jorgemoya: Fixed.

apps/web/vibes/soul/primitives/inline-email-form/index.tsx
jorgemoya: I agree but if so lets tackle in a different PR.

apps/web/vibes/soul/primitives/inline-email-form/index.tsx
jorgemoya: Updated to return `successMessage`.

apps/web/vibes/soul/examples/primitives/inline-email-form/actions.ts
```diff
@@ -15,7 +15,5 @@ export async function action(lastResult: SubmissionResult | null, formData: Form
   // Simulate a network request
   await new Promise((resolve) => setTimeout(resolve, 1000));
 
-  // const user = await logIn(submission.value)
-
-  return submission.reply({ resetForm: true });
+  return { ...submission.reply({ resetForm: true }), successMessage: 'Subscribed!' };
```
jorgemoya, line 18: What do you think of this approach, @migueloller?

apps/web/vibes/soul/primitives/inline-email-form/index.tsx
```diff
@@ -21,11 +22,13 @@ export function InlineEmailForm({
...
   submitLabel = 'Submit',
   placeholder = 'Enter your email',
+  dismissLabel,
 }: {
   className?: string;
   placeholder?: string;
   submitLabel?: string;
-  action: Action<SubmissionResult | null, FormData>;
+  dismissLabel?: string;
+  action: Action<(SubmissionResult & { successMessage?: string }) | null, FormData>;
```
jorgemoya, line 31: Update state type to include `successMessage`, but I wonder if we should create a new type that can be used in other components.

andrewreifman:
> I don't particularly like how the field shifts as well when there's an error. Seems like this needs some design love @andrewreifman

Agreed about the shift @migueloller. I think for that particular section we just need to align the items to the top instead of center. As far the as the red error text, you could be right in that it's hard to read, but I also wonder if the video makes it look worse. I need to test it out myself to see it, but maybe an easy solution is just tweaking the error color variable some as well.

jorgemoya:
Can we decide if we want to still use a toast for success or show a message like what @apledger is doing in the auth forms?

This is an example:

Screenshot 2024-12-06 at 2 27 52 PM

However, the form still shifts when these messages are shown, even if we top align. cc @andrewreifman @migueloller

andrewreifman:
> Can we decide if we want to still use a toast for success or show a message like what @apledger is doing in the auth forms?
> 
> This is an example:
> 
> Screenshot 2024-12-06 at 2 27 52 PM
> 
> However, the form still shifts when these messages are shown, even if we top align. cc @andrewreifman @migueloller

I'm good to use these messages in a container like the screenshot displays. Yeah, it will be impossible to avoid any shifting without absolute positioning, but you shouldn't really see 2 messages like that on top of each other in that section, right?

jorgemoya:
> I'm good to use these messages in a container like the screenshot displays. Yeah, it will be impossible to avoid any shifting without absolute positioning, but you shouldn't really see 2 messages like that on top of each other in that section, right?

Yeah, both statuses will never be shown at the same time. I updated the PR to use this instead of toast!

migueloller:
I like using those messages 👍🏻

How do we want to avoid the layout shift? Maybe we can anchor it below the input but absolute position? Or have a space there for it? The other forms should also have a space where the message goes to avoid the shift

apps/web/vibes/soul/examples/primitives/inline-email-form/actions.ts
```diff
@@ -15,7 +15,5 @@ export async function action(lastResult: SubmissionResult | null, formData: Form
   // Simulate a network request
   await new Promise((resolve) => setTimeout(resolve, 1000));
 
-  // const user = await logIn(submission.value)
-
-  return submission.reply({ resetForm: true });
+  return { ...submission.reply(), successMessage: 'Subscribed!' };
```
migueloller, line 18: We should do this instead:

```ts
return { lastResult: submission.reply(), successMessage: 'Subscribed!' }
```

apps/web/vibes/soul/primitives/inline-email-form/index.tsx
migueloller: See my other comment about wrapping the submission result instead

apps/web/vibes/soul/primitives/inline-email-form/index.tsx
```diff
@@ -74,8 +68,13 @@ export function InlineEmailForm({
...
       </div>
       {errors.map((error, index) => (
-        <FieldError key={index}>{error}</FieldError>
+        <FormStatus key={index} type="error">
+          {error}
+        </FormStatus>
       ))}
+      {form.status === 'success' && lastResult?.successMessage != null && (
+        <FormStatus>{lastResult.successMessage}</FormStatus>
+      )}
```
migueloller, line 77: Maybe for this form we do a toaster if we can't figure out a way to avoid the layout shift. We could also replace the form itself with the success message and match the height

migueloller (changes requested):
I think we shouldn't be spreading the submission result for the action. 

On a separate note, I was thinking what if we replaced the input with the success message to avoid the shift but we can't use that pattern for the error so perhaps using absolute position, or having an empty message that's got nothing in there but occupies the same space could help avoid the shift

apps/web/vibes/soul/primitives/inline-email-form/index.tsx
jorgemoya: Would you be okay if we track this design issue in a different issue?

migueloller: Yeah, let's do that. Can we please make a Linear issue so we don't lose track of it?

apps/web/vibes/soul/primitives/inline-email-form/index.tsx
```diff
@@ -25,9 +25,13 @@ export function InlineEmailForm({
...
   submitLabel?: string;
-  action: Action<SubmissionResult | null, FormData>;
+  action: Action<{ lastResult: SubmissionResult | null; successMessage?: string }, FormData>;
 }) {
-  const [lastResult, formAction, isPending] = useActionState(action, null);
+  const [{ lastResult, successMessage }, formAction, isPending] = useActionState(action, {
+    lastResult: null,
+  });
+
+  console.log('lastResult', lastResult);
```
migueloller, line 34: Stray `console.log`

migueloller (changes requested):
Looks good except for the stray `console.log` 😄

apps/web/vibes/soul/primitives/inline-email-form/index.tsx
jorgemoya: Fixed!

migueloller (approved):

PR #369, opened 2024-12-09
feat: add empty state to ProductsCarousel

hunterbecton:
## What/Why?

- Adds empty state to the `ProductsCarousel` component
- Simplifies examples for the `FeaturedProductsCarousel` by using a single component to handle loading and empty states
- Adds optional `loading` prop to the `ProductsCarousel` component to remove animation on the empty state
- Removes separate components for rendering `ProductsCarouselSkeleton` and `ProductsCarouselEmptyState`

https://github.com/user-attachments/assets/7394b0d2-a8a0-4f87-95fe-13ec3510154e

### DIFF
--- a/apps/web/vibes/soul/primitives/product-card/index.tsx
+++ b/apps/web/vibes/soul/primitives/product-card/index.tsx
@@ -1,4 +1,3 @@
-import { clsx } from 'clsx';
 import Image from 'next/image';
 import Link from 'next/link';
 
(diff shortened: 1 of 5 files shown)


apps/web/vibes/soul/primitives/products-carousel/index.tsx
```diff
@@ -20,75 +20,87 @@ export type CarouselProduct = CardProduct;
 interface Props {
   products: Streamable<CarouselProduct[]>;
   className?: string;
-  emptyStateMessage?: string;
+  emptyStateTitle?: string;
+  emptyStateDescription?: string;
```
migueloller, line 24: We've been using the pairs title/subtitle for situations like this. Can we rename `emptyStateDescription` to `emptyStateSubtitle`?

apps/web/vibes/soul/primitives/product-card/index.tsx
```diff
@@ -84,9 +84,15 @@ export function ProductCard({
...
+export function ProductCardSkeleton({
+  className,
+  loading = true,
+}: {
+  className?: string;
+  loading?: boolean;
+}) {
   return (
-    <div className={clsx('animate-pulse', className)}>
+    <div className={clsx(loading && 'animate-pulse', className)}>
```
migueloller, line 95: Not something we need to address in this PR but I pattern I saw in Catalyst `main` that I liked was using `data-pending` in various components and then using Tailwind `group` at the top level and then applying the class `group-has-[[data-pending]]:animate-pulse` at the top-level as well

migueloller: https://github.com/bigcommerce/catalyst/blob/d38f164d3e87ca87d3e792f8058a74c1f13e4220/core/app/%5Blocale%5D/(default)/(faceted)/search/page.tsx#L88

apps/web/vibes/soul/primitives/products-carousel/index.tsx
```diff
@@ -20,75 +20,87 @@ export type CarouselProduct = CardProduct;
...
+              <div className="absolute inset-0 mx-auto px-3 py-16 pb-3 @4xl:px-10 @4xl:pb-10 @4xl:pt-28">
+                <div className="mx-auto max-w-xl space-y-2 text-center @4xl:space-y-3">
+                  <h3 className="@4x:leading-none font-heading text-2xl leading-tight text-foreground @4xl:text-4xl">
+                    {emptyStateTitle}
+                  </h3>
+                  <p className="text-sm text-contrast-500 @4xl:text-lg">{emptyStateDescription}</p>
+                </div>
+              </div>
+            </Carousel>
+          );
```
migueloller, line 82: I think for the purpose of visual organization it might be nice to have this be in a separate component. Same goes for the Stream fallback (i.e., the pending state). Also, I think we want to export these empty and loading states as in certain cases that's useful to have for Catalyst either when rendering sections or when registering with Makeswift

pnpm-lock.yaml
migueloller, line 1: There's no changes to dependencies so I wouldn't expect lock file changes in this PR

migueloller (changes requested):
Looking good! There's some minor changes I think we should do

apps/web/vibes/soul/primitives/product-card/index.tsx
hunterbecton: I was having some issues getting `group-has` to work, but I included a different version that uses `group-data-[pending]`.

migueloller (approved):
Nice!

jorgemoya (approved):

migueloller (approved):

PR #370, opened 2024-12-09
feat: add dynamic form primitive

apledger:
Added a `DynamicForm` component to be reused and shared across all form components that are API driven. Such use cases include:

- Sign up form
- Address form
- Product details form


apps/web/vibes/soul/examples/primitives/form/action.ts
```diff
@@ -0,0 +1,111 @@
...
+  payload: FormData,
+) {
+  'use server';
+
+  const submission = parseWithZod(payload, { schema: schema(fields) });
+
+  if (submission.status !== 'success') {
+    return {
+      fields: prevState.fields,
+      lastResult: submission.reply(),
```
migueloller, line 101: Should this include a form error?

apps/web/vibes/soul/examples/primitives/form/action.ts
```diff
@@ -0,0 +1,111 @@
...
+    };
+  }
+
+  await new Promise((resolve) => setTimeout(resolve, 1000));
+
+  return {
+    fields: prevState.fields,
+    lastResult: submission.reply({}),
```
migueloller, line 109: Omit the empty object?

apps/web/vibes/soul/primitives/dynamic-form/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+  const defaultValue = fields.reduce<{
+    [Key in keyof SchemaRawShape]?: z.infer<SchemaRawShape[Key]>;
+  }>(
+    (acc, field) => ({
+      ...acc,
+      [field.name]: (syncQueryParams ? params[field.name] : null) ?? field.defaultValue ?? '',
+    }),
+    { quantity: 1 },
+  );
+  const [{ lastResult }, formAction] = useActionState(action, {
```
migueloller, line 69: If we've got the fields in the action state, should we rename the `field` prop to `initialFields` or `defaultFields` and allow the action to change the fields? For example, imagine an action that adds a new field to the form

apps/web/vibes/soul/primitives/dynamic-form/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+  field: Field;
+  formField: FieldMetadata<string | number | boolean | Date | undefined>;
+  syncQueryParams?: boolean;
+}) {
+  const controls = useInputControl(formField);
+  const [, setParam] = useQueryState(field.name, parseAsString.withOptions({ shallow: false }));
+
+  const handleChange = useCallback(
+    (value: string) => {
+      if (syncQueryParams === true) void setParam(value);
```
migueloller, line 134: I think we should use transitions to handle these async calls together with handling pending states

apps/web/vibes/soul/primitives/dynamic-form/schema.ts
migueloller, line 1: Have we tried defining a schema for each field and inferring the types from those instead of manually defining the types and then writing custom runtime code that attempts to emulate what Zod could do for us?

apps/web/vibes/soul/primitives/dynamic-form/index.tsx
migueloller, line 1: Our form fields look more like primitive inputs. What are we doing for field labels?

migueloller (commented):
I think this is a solid idea.

I wonder, though, if instead of a single component with a `fields` prop we should instead export all the different fields and let the user map the data? This would also solve the issue we have right now with having to figure out layout of fields that are side-by-side. What do you think?

apps/web/vibes/soul/primitives/dynamic-form/schema.ts
apledger: The problem is the type has to be driven by runtime value of `required`.

apps/web/vibes/soul/primitives/dynamic-form/index.tsx
apledger: Fixed.

apps/web/vibes/soul/examples/primitives/form/action.ts
apledger: Fixed

apps/web/vibes/soul/examples/primitives/form/action.ts
apledger: Fixed.

apps/web/vibes/soul/primitives/dynamic-form/index.tsx
apledger: Labels are included in the `Input` components. I want to separate out the `Input` and compose them into `Field` components that include a `label` and `error` prop.

migueloller: Ok, great. That's what I was going to suggest: to reserve the "field" naming for components that include labels and error messages and using "input" for primitives that don't include the labels or error messages

apps/web/vibes/soul/primitives/dynamic-form/schema.ts
migueloller: That should still be possible with Zod

apps/web/vibes/soul/examples/primitives/form/action.ts
migueloller: My bad, I think that `submission.reply()` includes field errors already via `parseWithZod`

apps/web/vibes/soul/form/swatch-radio-group/index.tsx
```diff
@@ -33,12 +33,11 @@ export const SwatchRadioGroup = React.forwardRef<
 >(({ id, label, options, errors, className, ...rest }, ref) => {
   return (
     <div className={clsx('space-y-2', className)}>
-      {label !== undefined && label !== '' && <Label htmlFor={id}>{label}</Label>}
+      {label !== undefined && label !== '' && <Label>{label}</Label>}
```
migueloller, line 36: We want to keep the `htmlFor` for accessibility. We could use `useId` instead of removing the prop

apps/web/vibes/soul/form/select/index.tsx
```diff
@@ -35,6 +34,8 @@ export function Select({
   value,
   ...rest
 }: Props) {
+  const id = React.useId();
```
migueloller, line 37: Change to `useId` and import from `'react'`

apps/web/vibes/soul/form/radio-group/index.tsx
```diff
@@ -57,3 +33,29 @@ export const RadioGroup = React.forwardRef<
 });
 
 RadioGroup.displayName = 'RadioGroup';
+
+function RadioGroupItem({ option, error = false }: { option: Option; error?: boolean }) {
+  const id = React.useId();
```
migueloller, line 38: Change to `useId` and import from `'react'`

apps/web/vibes/soul/form/radio-group/index.tsx
```diff
@@ -18,37 +18,13 @@ export const RadioGroup = React.forwardRef<
...
-              <RadioGroupPrimitive.Indicator className="relative flex size-full items-center justify-center after:block after:size-3 after:rounded-full after:bg-foreground" />
-            </RadioGroupPrimitive.Item>
-            <label className="pl-3 text-sm leading-none text-foreground" htmlFor={option.value}>
-              {option.label}
-            </label>
-          </div>
+      {label !== undefined && label !== '' && <Label>{label}</Label>}
+      <RadioGroupPrimitive.Root {...rest} aria-label={label} className="space-y-2" ref={ref}>
+        {options.map((option, index) => (
+          <RadioGroupItem error={errors != null && errors.length > 0} key={index} option={option} />
```
migueloller, line 27: What happened to using `value` as the `key`?

apps/web/vibes/soul/form/number-input/index.tsx
```diff
@@ -2,20 +2,22 @@
...
-  React.ComponentPropsWithoutRef<'input'> & {
+  Omit<React.ComponentPropsWithoutRef<'input'>, 'id'> & {
     label?: string;
     errors?: string[];
     decrementLabel?: string;
     incrementLabel?: string;
   }
->(({ id, label, className, required, errors, decrementLabel, incrementLabel, ...rest }, ref) => {
+>(({ label, className, required, errors, decrementLabel, incrementLabel, ...rest }, ref) => {
+  const id = React.useId();
```
migueloller, line 19: Change to `useId` and import from `'react'`

apps/web/vibes/soul/form/number-input/index.tsx
```diff
@@ -2,20 +2,22 @@
...
+import * as React from 'react';
 
 import { FieldError } from '@/vibes/soul/form/field-error';
 import { Label } from '@/vibes/soul/form/label';
 
-export const NumberInput = forwardRef<
+export const NumberInput = React.forwardRef<
   React.ComponentRef<'input'>,
-  React.ComponentPropsWithoutRef<'input'> & {
+  Omit<React.ComponentPropsWithoutRef<'input'>, 'id'> & {
```
migueloller, line 12: Weren't you planning on renaming these sort of component to, e.g., `NumberInputField`? If we do that, then having the props be all the input props is probably something we don't want to do. As it stands right now, though, we can probably continue to take in `id` and if it's not passed then we can default to `useId`

apps/web/vibes/soul/form/number-input/index.tsx
```diff
@@ -2,20 +2,22 @@
...
 import { clsx } from 'clsx';
 import { Minus, Plus } from 'lucide-react';
-import { forwardRef } from 'react';
+import * as React from 'react';
 
 import { FieldError } from '@/vibes/soul/form/field-error';
 import { Label } from '@/vibes/soul/form/label';
 
-export const NumberInput = forwardRef<
+export const NumberInput = React.forwardRef<
```
migueloller, line 10: How come the change to importing all of React instead of individual functions like we do elsewhere?

apps/web/vibes/soul/form/input/index.tsx
```diff
@@ -6,12 +6,14 @@ import { Label } from '@/vibes/soul/form/label';
...
   React.ComponentRef<'input'>,
-  React.ComponentPropsWithoutRef<'input'> & {
+  Omit<React.ComponentPropsWithoutRef<'input'>, 'id'> & {
     prepend?: React.ReactNode;
     label?: string;
     errors?: string[];
   }
->(({ id, prepend, label, className, required, errors, ...rest }, ref) => {
+>(({ prepend, label, className, required, errors, ...rest }, ref) => {
+  const id = React.useId();
```
migueloller, line 15: Change to `useId` and import from `'react'`

apps/web/vibes/soul/form/checkbox/index.tsx
```diff
@@ -8,15 +8,17 @@ import * as React from 'react';
...
 
-type Props = React.ComponentPropsWithoutRef<typeof CheckboxPrimitive.Root> & {
+type Props = Omit<React.ComponentPropsWithoutRef<typeof CheckboxPrimitive.Root>, 'id'> & {
   label?: React.ReactNode;
   errors?: string[];
 };
 
-export function Checkbox({ id, label, errors, className, ...rest }: Props) {
+export function Checkbox({ label, errors, className, ...rest }: Props) {
+  const id = React.useId();
```
migueloller, line 17: Change to `useId` and import from `'react'`

apps/web/vibes/soul/form/checkbox-group/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+  label,
+  options,
+  errors,
+  name,
+  value,
+  onValueChange,
+}: Props) {
+  return (
+    <div className={clsx('space-y-2', className)}>
+      {label !== undefined && label !== '' && <Label>{label}</Label>}
```
migueloller, line 37: Missing `htmlFor`

apps/web/vibes/soul/form/button-radio-group/index.tsx
```diff
@@ -21,12 +21,11 @@ export const ButtonRadioGroup = React.forwardRef<
 >(({ id, label, options, errors, className, ...rest }, ref) => {
   return (
     <div className={clsx('space-y-2', className)}>
-      {label !== undefined && label !== '' && <Label htmlFor={id}>{label}</Label>}
+      {label !== undefined && label !== '' && <Label>{label}</Label>}
```
migueloller, line 24: We should keep the `htmlFor` and use `useId`

apps/web/vibes/soul/primitives/dynamic-form/index.tsx
```diff
@@ -0,0 +1,351 @@
...
+                  })}
+                </div>
+              );
+            }
+
+            const formField = formFields[field.name];
+
+            if (formField == null) return null;
+
+            return <DynamicFormField field={field} formField={formField} key={field.name} />;
```
migueloller, line 118: Fo we not have a `formField.id` for these like we have above?

apps/web/vibes/soul/primitives/dynamic-form/schema.ts
```diff
@@ -0,0 +1,234 @@
...
+    | z.ZodOptional<z.ZodNumber>
+    | z.ZodArray<z.ZodString, 'atleastone' | 'many'>
+    | z.ZodOptional<z.ZodArray<z.ZodString, 'atleastone' | 'many'>>;
+
+  switch (field.type) {
+    case 'number':
+      fieldSchema = z.number();
+
+      if (field.min != null) fieldSchema.min(field.min);
+      if (field.max != null) fieldSchema.max(field.max);
```
migueloller, line 157: Both of these seem to be no-ops since we're not reassigning `fieldSchema`

apps/web/vibes/soul/primitives/dynamic-form/schema.ts
```diff
@@ -0,0 +1,234 @@
...
+  }[];
+} & FormField;
+
+type PasswordField = {
+  type: 'password';
+} & FormField;
+
+type ConfirmPasswordField = {
+  type: 'confirm-password';
+} & FormField;
```
migueloller, line 113: Is the reason for these two types of fields because the BigCommerce API has them and also to validate that one matches the other?

migueloller (changes requested):
This is looking pretty good! I really like this component ❤️ 

I have various comments and they're mostly nits. For the easy ones to fix, it would be great to do that. For the other ones that are more involved or require a refactor (e.g., renaming inputs to fields and extracting input components) I don't want to block progress on that as we can get to it after Jan 10.

There's also a couple more things that I'd like us to tackle after Jan 10 but they're mostly about consistency in code style, naming, etc. Things that matter more for the VIBES release and less so for Catalyst.

Nice work on this component, man!

apps/web/vibes/soul/form/checkbox-group/index.tsx
apledger: What is the label for in a group of inputs?

apps/web/vibes/soul/primitives/dynamic-form/schema.ts
apledger: Ah good catch

apps/web/vibes/soul/form/checkbox-group/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <div className="space-y-2">
+        {options.map((option) => (
+          <Checkbox
+            aria-label={option.label}
+            checked={value.includes(option.value)}
+            key={option.value}
+            label={option.label}
```
migueloller, line 44: Seems weird to pass both `aria-label` and `label` here. Seems like we should only need one of these, not both

apps/web/vibes/soul/primitives/dynamic-form/index.tsx
apledger: Not sure I understand this question

apps/web/vibes/soul/form/button-radio-group/index.tsx
apledger: There's no input directly correlated to this label since it's for the group.

apps/web/vibes/soul/primitives/dynamic-form/index.tsx
migueloller: Meant to say "Do we not". As in, we're using `field.name` as the React key, but above we have an ID that we can use. I was wondering if conform provided us with IDs for fields, if we could use that here as well

apps/web/vibes/soul/form/button-radio-group/index.tsx
migueloller: If there is no input then we might not want to use a `<label>` and instead have a label for the group as I describe in this comment: https://github.com/makeswift/vibes/pull/370#discussion_r1882329415

apps/web/vibes/soul/form/radio-group/index.tsx
apledger: There's no enforcement of value uniqueness in the API and I'd rather it not blow up and just show the duplicate option. AFAIK the main reason to not use index would be performance reasons right?

apps/web/vibes/soul/form/number-input/index.tsx
apledger: Not going to refactor this now, but good to know this is your preference. We have `import * as React` imports everywhere due to copying shadcn / radix.

apps/web/vibes/soul/form/checkbox-group/index.tsx
apledger: Done.

apps/web/vibes/soul/form/swatch-radio-group/index.tsx
apledger: Fixed.

apps/web/vibes/soul/form/checkbox-group/index.tsx
apledger: Fixed.

apps/web/vibes/soul/form/number-input/index.tsx
apledger: All of the form inputs are doing a `* as React` import. I'll go through later and make it all consistent.

apps/web/vibes/soul/form/number-input/index.tsx
apledger: I will refactor this later.

apps/web/vibes/soul/form/swatch-radio-group/index.tsx
```diff
@@ -30,16 +30,18 @@ export const SwatchRadioGroup = React.forwardRef<
     options: SwatchOption[];
     errors?: string[];
   }
->(({ id, label, options, errors, className, ...rest }, ref) => {
+>(({ label, options, errors, className, ...rest }, ref) => {
+  const id = React.useId();
+
   return (
     <div className={clsx('space-y-2', className)}>
       {label !== undefined && label !== '' && <Label htmlFor={id}>{label}</Label>}
```
migueloller, line 38: We should use `id`, not `htmlFor` when using `aria-labelledby` (this seems to be needed across a couple of components)

migueloller (dismissed):

apps/web/vibes/soul/form/swatch-radio-group/index.tsx
apledger: Done.

migueloller (approved):

jorgemoya (approved):

PR #371, opened 2024-12-10
feat: make CursorPagination component accessible

migueloller:
The changes here are heavily inspired by shadcn/ui Pagination component.

See https://ui.shadcn.com/docs/components/pagination.

https://github.com/user-attachments/assets/0a9712e4-05a8-4d97-ad93-c3d48922b35d

### DIFF
--- a/apps/web/vibes/soul/primitives/cursor-pagination/index.tsx
+++ b/apps/web/vibes/soul/primitives/cursor-pagination/index.tsx
@@ -5,7 +5,9 @@ import { ArrowLeft, ArrowRight } from 'lucide-react';
 import Link from 'next/link';
 import { useSearchParams } from 'next/navigation';
 import { createSerializer, parseAsString } from 'nuqs';
-import { Suspense, use } from 'react';
+import { Suspense } from 'react';
+
+import { Streamable, useStreamable } from '@/vibes/soul/lib/streamable';
 
 export interface CursorPaginationInfo {
   startCursorParamName?: string;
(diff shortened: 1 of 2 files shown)


apps/web/vibes/soul/primitives/cursor-pagination/index.tsx
```diff
@@ -15,7 +17,10 @@ export interface CursorPaginationInfo {
 }
 
 interface Props {
-  info: CursorPaginationInfo | Promise<CursorPaginationInfo>;
+  label?: Streamable<string | null>;
+  info: Streamable<CursorPaginationInfo>;
+  previousLabel?: Streamable<string | null>;
+  nextLabel?: Streamable<string | null>;
```
jorgemoya, line 23: Do you have an example of why these label values are streamable?

migueloller: It's because these are translated and getting translations is asynchronous

migueloller: As I was implementing changes to the category pages in Catalyst I realized that labels need to be streamable for this reason

jorgemoya: So do we need to make all labels Streamable?

migueloller: Yeah, probably. I'm planning on doing this as-needed as I work on Catalyst and VIBES

jorgemoya (approved):

PR #381, opened 2024-12-11
fix: add logoHref and logoLabel props to footer

jorgemoya:
Add props for accessibility. Remove link when not being used.

### DIFF
--- a/apps/web/vibes/soul/sections/footer/index.tsx
+++ b/apps/web/vibes/soul/sections/footer/index.tsx
@@ -38,6 +38,8 @@ interface Props {
   paymentIcons?: Streamable<ReactNode[] | null>;
   socialMediaLinks?: Streamable<SocialMediaLink[] | null>;
   className?: string;
+  logoHref?: string;
+  logoLabel?: string;
 }
 
 export const Footer = forwardRef(function Footer(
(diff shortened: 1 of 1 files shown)


migueloller (commented):
What are the accessibility issues? It seems this component was simply unfinished. We should have the href as a prop as well as a label for it. It seems common to have a link back to the home page in the header ir footer, though

jorgemoya:
> What are the accessibility issues? It seems this component was simply unfinished. We should have the href as a prop as well as a label for it. It seems common to have a link back to the home page in the header ir footer, though

Links need an accessible label.

I disagree regarding needing this link. Lets discuss offline.

jorgemoya:
Updated PR to keep Link but don't use if not needed.

apps/web/vibes/soul/sections/footer/index.tsx
```diff
@@ -74,22 +78,30 @@ export const Footer = forwardRef(function Footer(
...
                       <span className="whitespace-nowrap font-heading text-2xl font-semibold">
                         {logo}
                       </span>
                     </Link>
+                  ) : (
+                    <div className="relative mb-2 inline-block h-10 w-full max-w-56 rounded-lg ring-primary focus-visible:outline-0 focus-visible:ring-2">
+                      <span className="whitespace-nowrap font-heading text-2xl font-semibold">
+                        {logo}
+                      </span>
+                    </div>
```
migueloller, line 96: What if instead of duplicating the content we default the `href` to `#`?

jorgemoya: Fixed on latest commit.

migueloller (approved):

PR #383, opened 2024-12-11
feat: add empty state to product list section

hunterbecton:
## What/Why
- Adds `count` prop to `ProductList` so sections that use this component can render a custom number of empty state cards 
- Adds `emptyStateTitle` and `emptyStateDescription` to `ProductsListSection`

## Testing
https://github.com/user-attachments/assets/bdfb2cb7-ecc0-4377-a30d-837e3528104f

### DIFF
--- a/apps/web/vibes/soul/primitives/products-carousel/index.tsx
+++ b/apps/web/vibes/soul/primitives/products-carousel/index.tsx
@@ -19,11 +19,12 @@ export type CarouselProduct = CardProduct;
 interface Props {
   products: Streamable<CarouselProduct[]>;
   className?: string;
-  emptyStateTitle?: string;
-  emptyStateSubtitle?: string;
+  emptyStateTitle?: Streamable<string | null>;
+  emptyStateSubtitle?: Streamable<string | null>;
   scrollbarLabel?: string;
   previousLabel?: string;
   nextLabel?: string;
+  placeholderCount?: number;
 }
 
 export function ProductsCarousel({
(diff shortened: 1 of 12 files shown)


apps/web/vibes/soul/sections/products-list-section/index.tsx
```diff
@@ -98,6 +98,9 @@ export function ProductsListSection({
                 compareLabel={compareLabel}
                 compareParamName={compareParamName}
                 compareProducts={compareProducts}
+                count={8}
+                emptyStateSubtitle="Change your filters to see more products"
+                emptyStateTitle="No products found"
```
migueloller, line 103: These need to be streamable props so that they can be localized

apps/web/vibes/soul/primitives/products-list/index.tsx
```diff
@@ -21,6 +21,7 @@ interface Props {
   compareParamName?: string;
   emptyStateTitle?: string;
   emptyStateSubtitle?: string;
+  count?: number;
```
migueloller, line 24: We should rename this to `skeletonCardCount`

Taking a step back, though, should this even be configurable?

migueloller (changes requested):

apps/web/vibes/soul/primitives/products-list/index.tsx
hunterbecton: You know, this value does not need to change. I can't find an example right now where it makes sense to make it configurable.

migueloller: I agree. I think we can keep it non-configurable for now

apps/web/vibes/soul/primitives/products-carousel/index.tsx
```diff
@@ -19,11 +19,12 @@ export type CarouselProduct = CardProduct;
 interface Props {
   products: Streamable<CarouselProduct[]>;
   className?: string;
-  emptyStateTitle?: string;
-  emptyStateSubtitle?: string;
+  emptyStateTitle?: Streamable<string | null>;
```
hunterbecton, line 22: @migueloller Here's where I need clarification. In our discussion on Slack, you mentioned that primitives would get normal values in their props, but if this prop gets past the value from its parent, wouldn't it also need to be Streamable?

migueloller (approved):

apps/web/vibes/soul/primitives/products-carousel/index.tsx
migueloller: This primitive is a bit more section-y than others. It handles its own skeleton, etc. so it makes sense for it to take streamable values. It's more about whether it makes sense or not for the component to have a fallback. For example, a skeleton shouldn't take a streamable value because what would the fallback for a skeleton be?

PR #384, opened 2024-12-11
feat: add streamable props to ProductDescription

jorgemoya:
- Add streamable accordions and image props.
- Suspense and render skeleton.
- Update examples.

https://github.com/user-attachments/assets/dc99cf1c-1c23-4ee3-88f1-76c4981f2ede

### DIFF
--- a/apps/web/vibes/soul/sections/product-description/index.tsx
+++ b/apps/web/vibes/soul/sections/product-description/index.tsx
@@ -9,7 +9,7 @@ export interface AccordionItem {
 }
 
 interface Props {
-  accordions: Streamable<AccordionItem[] | null>;
+  accordions: Streamable<AccordionItem[]>;
   image?: Streamable<{
     src: string;
     alt: string;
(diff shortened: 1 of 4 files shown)


apps/web/vibes/soul/sections/product-description/index.tsx
```diff
@@ -8,41 +9,83 @@ export interface AccordionItem {
 }
 
 interface Props {
-  accordions: AccordionItem[];
-  image?: {
+  accordions: Streamable<AccordionItem[] | null>;
```
jorgemoya, line 12: Actually, do arrays still need `null`? 🤔

migueloller: Yeah, I think we should do it to be consistent. That being said, the `accordions` prop is required so for that reason we shouldn't have `null` here

migueloller (dismissed):
LGTM! For the accordion skeleton, is there a reason we're using a dot for the arrow instead of an arrow? Seems like this should stay the same? Also, do we want to default to the first accordion in the skeleton being open or perhaps closed? Or maybe we can make this configurable?

jorgemoya:
> LGTM! For the accordion skeleton, is there a reason we're using a dot for the arrow instead of an arrow? Seems like this should stay the same? Also, do we want to default to the first accordion in the skeleton being open or perhaps closed? Or maybe we can make this configurable?

For design questions I would defer to @andrewreifman .

I think we should default to the first item being open, seems weird to have all closed initially. Maybe it can be configured which one is default opened and if we should default open.

migueloller:
> > LGTM! For the accordion skeleton, is there a reason we're using a dot for the arrow instead of an arrow? Seems like this should stay the same? Also, do we want to default to the first accordion in the skeleton being open or perhaps closed? Or maybe we can make this configurable?
> 
> For design questions I would defer to @andrewreifman .
> 
> I think we should default to the first item being open, seems weird to have all closed initially. Maybe it can be configured which one is default opened and if we should default open.

Yeah, I agree that we should default to having some of them open. And perhaps the skeleton should be configurable too. Don't want that to block this PR, though. What you have looks good

PR #385, opened 2024-12-11
feat: add streamable props to ProductDetail

jorgemoya:
+ Adds streamable props
+ Renders skeleton when loading
+ Updates examples

For this, I'm not sure if the best way to go is to add the streamable prop to each of the inner components or the way I implemented this the right solution. I also understand if we want to move the skeletons to the inner components. Let me know. cc @migueloller 

https://github.com/user-attachments/assets/fc57aca2-b948-4590-a59d-c0f1cd1b7ab2


migueloller (approved):
This looks good! I think it's quite reasonable to keep streaming to sections. We could push it down to primitives but it's something we can assess later as-needed

PR #389, opened 2024-12-13
feat: add accessible props to ProductDetail

jorgemoya:
Adds `quantityLabel` prop to ProductDetail that is then forwarded to form and set as an `aria-label` to the quantity field since we don't have a visible label.

Screenshot 2024-12-12 at 7 28 27 PM


jorgemoya:
Closing #338 in favor of this PR.

migueloller (approved):

PR #391, opened 2024-12-13
feat: add prefetching to product detail form

jorgemoya:
This change allows prefetching to happen when multi select form elements are moused over.  This feature can be enabled by passing `prefetch=true`.

https://github.com/user-attachments/assets/f7625fc0-fffa-4f84-864e-b3ac0857b689

### DIFF
--- a/apps/web/vibes/soul/form/button-radio-group/index.tsx
+++ b/apps/web/vibes/soul/form/button-radio-group/index.tsx
@@ -17,8 +17,9 @@ export const ButtonRadioGroup = React.forwardRef<
     label?: string;
     options: Option[];
     errors?: string[];
+    onOptionMouseEnter?: (value: string) => void;
   }
->(({ label, options, errors, className, ...rest }, ref) => {
+>(({ label, options, errors, className, onOptionMouseEnter, ...rest }, ref) => {
   const id = React.useId();
 
   return (
@@ -43,6 +44,11 @@ export const ButtonRadioGroup = React.forwardRef<
             disabled={option.disabled}
             id={option.value}
             key={option.value}
+            onMouseEnter={() => {
+              if (typeof onOptionMouseEnter === 'function') {
+                onOptionMouseEnter(option.value);
+              }
+            }}
             value={option.value}
           >
             {option.label}
--- a/apps/web/vibes/soul/form/button-radio-group/index.tsx
+++ b/apps/web/vibes/soul/form/button-radio-group/index.tsx
@@ -45,9 +45,7 @@ export const ButtonRadioGroup = React.forwardRef<
             id={option.value}
             key={option.value}
             onMouseEnter={() => {
-              if (typeof onOptionMouseEnter === 'function') {
-                onOptionMouseEnter(option.value);
-              }
+              onOptionMouseEnter?.(option.value);
             }}
             value={option.value}
           >
--- a/apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
+++ b/apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
@@ -9,6 +9,7 @@ import {
   useInputControl,
 } from '@conform-to/react';
 import { getZodConstraint, parseWithZod } from '@conform-to/zod';
+import { usePathname, useRouter, useSearchParams } from 'next/navigation';
 import { parseAsString, useQueryState, useQueryStates } from 'nuqs';
 import { ReactNode, useActionState, useCallback, useEffect } from 'react';
 import { useFormStatus } from 'react-dom';
@@ -48,6 +49,7 @@ interface Props<F extends Field> {
   incrementLabel?: string;
   decrementLabel?: string;
   ctaDisabled?: boolean;
+  prefetch?: boolean;
 }
 
 export function ProductDetailForm<F extends Field>({
@@ -59,6 +61,7 @@ export function ProductDetailForm<F extends Field>({
   incrementLabel = 'Increase quantity',
   decrementLabel = 'Decrease quantity',
   ctaDisabled = false,
+  prefetch = false,
 }: Props<F>) {
   const [params] = useQueryStates(
     fields.reduce<Record<string, typeof parseAsString>>(
@@ -118,6 +121,7 @@ export function ProductDetailForm<F extends Field>({
                 formField={formFields[field.name]!}
                 // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
                 key={formFields[field.name]!.id}
+                prefetch={prefetch}
               />
             );
           })}
@@ -166,10 +170,16 @@ function SubmitButton({ children, disabled }: { children: React.ReactNode; disab
 function FormField({
   field,
   formField,
+  prefetch,
 }: {
   field: Field;
   formField: FieldMetadata<string | number | boolean | Date | undefined>;
+  prefetch: boolean;
 }) {
+  const router = useRouter();
+  const pathname = usePathname();
+  const searchParams = useSearchParams();
+
   const controls = useInputControl(formField);
   const [, setParam] = useQueryState(field.name, parseAsString.withOptions({ shallow: false }));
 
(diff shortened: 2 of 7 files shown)


jorgemoya:
@migueloller Any chance you missed this?

migueloller:
> @migueloller Any chance you missed this?

I didn't. I just didn't get to it. Reviewing now!

apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
```diff
@@ -154,6 +164,18 @@ function FormField({
...
 
+  const handleOnOptionMouseEnter = (value: string) => {
+    if (prefetch) {
+      const optionSearchParams = new URLSearchParams(searchParams.toString());
+
+      optionSearchParams.set(String(field.name), String(value));
+
+      const newUrl = `${pathname}?${optionSearchParams.toString()}`;
+
+      router.prefetch(newUrl);
```
migueloller, line 175: Is the idea here that we're doing what will happen when an option is selected on mouse enter? If so, we shouldn't use `URLSearchParams` and should use `nuqs` to keep things consistent (i.e., use the same parsers, setters, etc.)

apps/web/vibes/soul/form/button-radio-group/index.tsx
```diff
@@ -43,6 +44,11 @@ export const ButtonRadioGroup = React.forwardRef<
             disabled={option.disabled}
             id={option.value}
             key={option.value}
+            onMouseEnter={() => {
+              if (typeof onOptionMouseEnter === 'function') {
+                onOptionMouseEnter(option.value);
+              }
+            }}
```
migueloller, line 51:
```suggestion
            onMouseEnter={() => {
              onOptionMouseEnter?.(option.value);
            }}
```

I suggest we apply that change to all instances of this pattern

migueloller (changes requested):
I'm ambivalent to doing something like this.

It feels a bit weird how we're implementing an `onOptionMouseEnter` that doesn't pass a `MouseEvent` back. Perhaps the solution there is to refactor the components to be more "composable" so that we can render the individual items and then we can pass `onMouseEnter` to the specific option. That's something we don't want to spend time on right now, though.

Something else that feels off to me is that while we persist the form's fields in the URL, this doesn't necessarily mean that the user of this section will use those same params for the routing. Is that not the case? It feels like we might be coupling Catalyst's routing to VIBES. Maybe I'm overthinking things here?

Regardless, I think we should handle all URL search params using `nuqs`. If we make that slight change, we can move forward with this. We can always come back and change it if we change our minds.

apps/web/vibes/soul/form/button-radio-group/index.tsx
jorgemoya: I was trying to be more type safe but this is fine by me.

apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
jorgemoya: Not sure I follow. We don't want to set params here, we only want to know the url we need to prefetch. Setting params in nuqs would trigger a redirect no?

jorgemoya:
> I'm ambivalent to doing something like this.
> 
> It feels a bit weird how we're implementing an `onOptionMouseEnter` that doesn't pass a `MouseEvent` back. Perhaps the solution there is to refactor the components to be more "composable" so that we can render the individual items and then we can pass `onMouseEnter` to the specific option. That's something we don't want to spend time on right now, though.

Yes, this could work better, but yes would be a refactor that we might not want to do right now.

> 
> Something else that feels off to me is that while we persist the form's fields in the URL, this doesn't necessarily mean that the user of this section will use those same params for the routing. Is that not the case? It feels like we might be coupling Catalyst's routing to VIBES. Maybe I'm overthinking things here?
> 

I don't think this it the case. We can see that we set `field.name` as the param name, which is what we use to generate the url we want to prefetch. Unless I'm misunderstanding.

`const [, setParam] = useQueryState(field.name, parseAsString.withOptions({ shallow: false }));`

> Regardless, I think we should handle all URL search params using `nuqs`. If we make that slight change, we can move forward with this. We can always come back and change it if we change our minds.

Since we are not setting a url, can we really use `nuqs`?

apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
migueloller: `nuqs` provides serializers to turn values into URL search params that can share the same parsers we're using in `useQueryState`. More detail [here](https://nuqs.47ng.com/docs/utilities#serializer-helper)

migueloller:
> > I'm ambivalent to doing something like this.
> > It feels a bit weird how we're implementing an `onOptionMouseEnter` that doesn't pass a `MouseEvent` back. Perhaps the solution there is to refactor the components to be more "composable" so that we can render the individual items and then we can pass `onMouseEnter` to the specific option. That's something we don't want to spend time on right now, though.
> 
> Yes, this could work better, but yes would be a refactor that we might not want to do right now. What I could do is change the name of the property? So that it is clear that this is not the event handler, but a function that is called within the event? `prefetchHandler`?
> 

I think what you have right now is good. `prefetchHandler` might couple it too much for the low-level primitive. We can revisit later.

> > Something else that feels off to me is that while we persist the form's fields in the URL, this doesn't necessarily mean that the user of this section will use those same params for the routing. Is that not the case? It feels like we might be coupling Catalyst's routing to VIBES. Maybe I'm overthinking things here?
> 
> I don't think this it the case. We can see that we set `field.name` as the param name, which is what we use to generate the url we want to prefetch. Unless I'm misunderstanding.
> 
> `const [, setParam] = useQueryState(field.name, parseAsString.withOptions({ shallow: false }));`
> 

I guess you're right. The intent behind serializing state in URL params with `nuqs` was for deep linking. If we're assuming that the deep link works, then preloading at that URL should work too 👍🏻 

> > Regardless, I think we should handle all URL search params using `nuqs`. If we make that slight change, we can move forward with this. We can always come back and change it if we change our minds.
> 
> Since we are not setting a url, can we really use `nuqs`?

https://github.com/makeswift/vibes/pull/391#discussion_r1887607666

apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
jorgemoya: Updated to use nuqs 👍

apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
```diff
@@ -166,10 +169,16 @@ function SubmitButton({ children, disabled }: { children: React.ReactNode; disab
...
   formField,
+  prefetch,
 }: {
   field: Field;
   formField: FieldMetadata<string | number | boolean | Date | undefined>;
+  prefetch: boolean;
 }) {
+  const router = useRouter();
+  const pathname = usePathname();
+  const searchParams = useSearchParams();
```
migueloller, line 180: We still seem to be using `useSearchParams` . Can't we use the result of `useQueryState` for this instead of `useSearchParams`?

I guess we're trying to get _all_ search params here, not just the ones being managed by `nuqs`?

migueloller (dismissed):

apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
jorgemoya: Yes, if I'm not wrong, I believe nuqs only manages the state param of this specific option, no?

From my testing, I couldn't get all params from this state we have in the field, unless there's another utility I'm missing?

Ultimately we want to get all params and just change the one for this field to the new option.

migueloller: Yeah, `nuqs` will only retrieve and change the params that have parsers defined. It does this so it doesn't clobber other search params that could have been set by something else, like pagination params or UTM params. It's a very nice design but you have to be aware of it. For example, in the PLP I had to handle pagination params in a special way so that changing PLP filters cleared pagination params. This meant I had to have `nuqs` "track" them so that they could be cleared when I called `setParams(null)`

migueloller (dismissed):

migueloller (approved):

PR #392, opened 2024-12-13
feat: show toasts on successfully adding a product

jorgemoya:
This will show a toast of a simple string when product is added successfully. We still need to think about how to pass in string interpolation so that we can render a more custom message with a link that understand the quantity added.

Screenshot 2024-12-13 at 2 02 07 PM

### DIFF
--- a/apps/web/vibes/soul/primitives/alert/index.tsx
+++ b/apps/web/vibes/soul/primitives/alert/index.tsx
@@ -1,11 +1,12 @@
 import { clsx } from 'clsx';
 import { X } from 'lucide-react';
+import { ReactNode } from 'react';
 
 import { Button } from '@/vibes/soul/primitives/button';
 
 interface Props {
   variant: 'success' | 'warning' | 'error' | 'info';
-  message: string;
+  message: ReactNode;
   description?: string;
   dismissLabel?: string;
   action?: {
--- a/apps/web/vibes/soul/primitives/toaster/index.tsx
+++ b/apps/web/vibes/soul/primitives/toaster/index.tsx
@@ -1,5 +1,6 @@
 'use client';
 
+import { ReactNode } from 'react';
 import { Toaster as Sonner, toast as SonnerToast } from 'sonner';
 
 import { Alert } from '@/vibes/soul/primitives/alert';
@@ -31,7 +32,7 @@ export const Toaster = ({ ...props }: ToasterProps) => {
 };
 
 export const toast = {
-  success: (message: string, options?: ToastOptions) => {
+  success: (message: ReactNode, options?: ToastOptions) => {
     const position = options?.position;
 
     const toastId = SonnerToast(
@@ -44,7 +45,7 @@ export const toast = {
       { position },
     );
   },
-  error: (message: string, options?: ToastOptions) => {
+  error: (message: ReactNode, options?: ToastOptions) => {
     const position = options?.position;
 
     const toastId = SonnerToast(
@@ -57,7 +58,7 @@ export const toast = {
       { position },
     );
   },
-  warning: (message: string, options?: ToastOptions) => {
+  warning: (message: ReactNode, options?: ToastOptions) => {
     const position = options?.position;
 
     const toastId = SonnerToast(
@@ -70,7 +71,7 @@ export const toast = {
       { position },
     );
   },
-  info: (message: string, options?: ToastOptions) => {
+  info: (message: ReactNode, options?: ToastOptions) => {
     const position = options?.position;
 
     const toastId = SonnerToast(
--- a/apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
+++ b/apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
@@ -10,7 +10,7 @@ import {
 } from '@conform-to/react';
 import { getZodConstraint, parseWithZod } from '@conform-to/zod';
 import { parseAsString, useQueryState, useQueryStates } from 'nuqs';
-import { useActionState, useCallback } from 'react';
+import { useActionState, useCallback, useEffect } from 'react';
 import { useFormStatus } from 'react-dom';
 import { z } from 'zod';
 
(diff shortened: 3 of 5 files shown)


apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
```diff
@@ -72,10 +76,18 @@ export function ProductDetailForm<F extends Field>({
...
   const [{ lastResult }, formAction] = useActionState(action, {
     fields,
     lastResult: null,
   });
+
+  useEffect(() => {
+    if (lastResult?.status === 'success') {
+      toast.success(successMessage);
+    }
+  }, [lastResult, successMessage]);
```
migueloller, line 89: We should do this `onSubmit`, not on `useEffect`

migueloller (changes requested):

apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
jorgemoya: You're right! 🤦

jorgemoya: Question: what is the best way?

`onSubmit` in `useForm` is called before server validation.

migueloller: That's a good question. What about something like this?

```tsx
function SomeForm({ action }) {
  const [{ lastResult }, formAction, isPending] = useActionState(action, { lastResult: null })
  const [form] = useForm({
    lastResult,
    onSubmit(event, formData) {
      // prevent `formAction` from being called, we'll call `action` ourselves
      event.preventDefault()

      startTransition(async () => {
        const payload = await action(lastResult, formData)

        await toaster.success(payload.successMessage)
      })
    }
  })

  return <form action={formAction} onSubmit={form.onSubmit} />
}
```

This has problems with keeping the action state up to date since we're not calling `formAction`. Ideally we can call `formAction` and await that but I'm not sure that's possible.

We could also pass a client action to `useActionState` and that would work well. It would mean we don't support progressive enhancement, but that's probably fine.

```tsx
function SomeForm({ action }) {
  const [{ lastResult }, formAction, isPending] = useActionState(async (prevState, formData) => {
    const nextState = await action(lastResult, formData)

    await toaster.success(nextState.successMessage)

    return nextState
  }, { lastResult: null })
  const [form] = useForm({ lastResult })

  return <form action={formAction} onSubmit={form.onSubmit} />
}
```

migueloller: Maybe let's just continue using `useEffect` for now?

jorgemoya: I implemented a solution that works but has a smell, let me know what you think.

apps/web/vibes/soul/primitives/alert/index.tsx
```diff
@@ -1,11 +1,12 @@
 import { clsx } from 'clsx';
 import { X } from 'lucide-react';
+import { ReactNode } from 'react';
 
 import { Button } from '@/vibes/soul/primitives/button';
 
 interface Props {
   variant: 'success' | 'warning' | 'error' | 'info';
-  message: string;
+  message: ReactNode;
```
jorgemoya, line 9: Changed to `ReactNode` because we want to be able to toast messages that include `Link` components.

apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
```diff
@@ -86,7 +91,26 @@ export function ProductDetailForm<F extends Field>({
...
+        const payload = await action({ fields, lastResult }, formData);
+
+        console.log(payload);
+
+        if (payload.lastResult?.status === 'success') {
+          toast.success(payload.successMessage);
+        }
+
+        if (payload.lastResult?.status === 'error') {
+          toast.error(payload.lastResult.error?.['']);
```
jorgemoya, line 108: This seems a little off, but not sure if this is the expected pattern when an error message is not tied to a specific id.

apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
jorgemoya: Reverted to `useEffect`.

migueloller: I think using `useEffect` is fine for now

jorgemoya:
Handles error now:

Screenshot 2024-12-16 at 2 29 33 PM

apps/web/vibes/soul/primitives/toaster/index.tsx
```diff
@@ -31,7 +32,7 @@ export const Toaster = ({ ...props }: ToasterProps) => {
 };
 
 export const toast = {
-  success: (message: string, options?: ToastOptions) => {
+  success: (message: ReactNode, options?: ToastOptions) => {
```
migueloller, line 35: For all these builtin toast methods, wouldn't we want to use something like

```tsx
toast.custom(<Alert type="error" message={<p>Message with <Link href="#">link</Link>.} />)
```

apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
```diff
@@ -105,6 +119,11 @@ export function ProductDetailForm<F extends Field>({
               />
             );
           })}
+          {form.errors?.map((error, index) => (
+            <FormStatus className="pt-3" key={index} type="error">
```
migueloller, line 123: 🍹 Won't a single error appear too far from the form? Feels weird to use padding here. Might be better to wrap the errors in a `<div>` with some `gap`? Or has this been what we've been doing for all forms?

migueloller (approved):
LGTM 👍🏻

apps/web/vibes/soul/primitives/toaster/index.tsx
jorgemoya: This is not needed, unless I'm misunderstanding.

Screenshot 2024-12-16 at 3 02 04 PM

apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
jorgemoya: Attached a screenshot above:

Screenshot 2024-12-16 at 2 29 33 PM

What you suggest makes sense, but we've been doing it like this all forms. We should clean this up in a different PR and fix all of them.

PR #394, opened 2024-12-13
feat: stream in submit button label and disabled status

jorgemoya:
This allows disabling the button from an API responses as well as setting the label depending on the status.

Screenshot 2024-12-13 at 2 53 01 PM

### DIFF
--- a/apps/web/vibes/soul/sections/product-detail/index.tsx
+++ b/apps/web/vibes/soul/sections/product-detail/index.tsx
@@ -27,8 +27,8 @@ interface Props<F extends Field> {
   quantityLabel?: string;
   incrementLabel?: string;
   decrementLabel?: string;
-  submitButtonLabel?: Streamable<string>;
-  submitButtonDisabled?: Streamable<boolean>;
+  ctaLabel?: Streamable<string | null>;
+  ctaDisabled?: Streamable<boolean | null>;
 }
 
 export function ProductDetail<F extends Field>({
@@ -39,8 +39,8 @@ export function ProductDetail<F extends Field>({
   quantityLabel,
   incrementLabel,
   decrementLabel,
-  submitButtonLabel: streamableSubmitButtonLabel,
-  submitButtonDisabled: streamableSubmitButtonDisabled,
+  ctaLabel: streamableCtaLabel,
+  ctaDisabled: streamableCtaDisabled,
 }: Props<F>) {
   return (
     <section className="@container">
@@ -103,16 +103,16 @@ export function ProductDetail<F extends Field>({
                     fallback={<ProductDetailFormSkeleton />}
                     value={Promise.all([
                       streamableFields,
-                      streamableSubmitButtonLabel,
-                      streamableSubmitButtonDisabled,
+                      streamableCtaLabel,
+                      streamableCtaDisabled,
                     ])}
                   >
-                    {([fields, submitButtonLabel, submitButtonDisabled]) => (
+                    {([fields, ctaLabel, ctaDisabled]) => (
                       <ProductDetailForm
                         action={action}
-                        ctaLabel={submitButtonLabel}
+                        ctaLabel={ctaLabel ?? undefined}
                         decrementLabel={decrementLabel}
-                        disabled={submitButtonDisabled}
+                        ctaDisabled={ctaDisabled ?? undefined}
                         fields={fields}
                         incrementLabel={incrementLabel}
                         productId={product.id}
--- a/apps/web/vibes/soul/sections/product-detail/index.tsx
+++ b/apps/web/vibes/soul/sections/product-detail/index.tsx
@@ -24,10 +24,10 @@ interface Props<F extends Field> {
   product: Streamable<ProductDetailProduct | null>;
   action: ProductDetailFormAction<F>;
   fields: Streamable<F[]>;
-  ctaLabel?: string;
   quantityLabel?: string;
   incrementLabel?: string;
   decrementLabel?: string;
+  submitButton?: Streamable<{ label: string; disabled?: boolean }>;
 }
 
 export function ProductDetail<F extends Field>({
@@ -35,10 +35,10 @@ export function ProductDetail<F extends Field>({
   action,
   fields: streamableFields,
   breadcrumbs: streamableBreadcrumbs,
-  ctaLabel,
   quantityLabel,
   incrementLabel,
   decrementLabel,
+  submitButton: streamableSubmitButton,
 }: Props<F>) {
   return (
     <section className="@container">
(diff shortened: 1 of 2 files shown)


apps/web/vibes/soul/sections/product-detail/index.tsx
```diff
@@ -99,15 +99,23 @@ export function ProductDetail<F extends Field>({
...
-                        decrementLabel={decrementLabel}
-                        fields={fields}
-                        incrementLabel={incrementLabel}
-                        productId={product.id}
-                        quantityLabel={quantityLabel}
-                      />
+                      <Stream
+                        fallback={<ProductDetailFormSkeleton />}
+                        value={streamableSubmitButton}
+                      >
```
migueloller, line 105: This feels off to me. We've got way to many uses of `<Stream>` and granular skeletons. We should combine streamable with `Promise.all` and use a single `<Stream>`

apps/web/vibes/soul/sections/product-detail/index.tsx
```diff
@@ -24,21 +24,21 @@ interface Props<F extends Field> {
   product: Streamable<ProductDetailProduct | null>;
   action: ProductDetailFormAction<F>;
   fields: Streamable<F[]>;
-  ctaLabel?: string;
   quantityLabel?: string;
   incrementLabel?: string;
   decrementLabel?: string;
+  submitButton?: Streamable<{ label: string; disabled?: boolean }>;
```
migueloller, line 30: How come we're grouping in an object? We haven't been doing this for other labels. Also, shouldn't the submit button's disabled state be handled internally when the form is being submitted, etc.?

In what instance would we want to pass this value in as a hardcoded `false` value?

apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
```diff
@@ -118,19 +121,25 @@ export function ProductDetailForm<F extends Field>({
...
 
-function SubmitButton({ children }: { children: React.ReactNode }) {
+function SubmitButton({ children, disabled }: { children: React.ReactNode; disabled?: boolean }) {
   const { pending } = useFormStatus();
 
   return (
-    <Button className="w-auto @xl:w-56" loading={pending} size="medium" type="submit">
+    <Button
+      className="w-auto @xl:w-56"
+      disabled={disabled}
```
migueloller, line 138: This should probably be disabled while `pending`. Does the `loading` prop not take care of that?

apps/web/vibes/soul/sections/product-detail/index.tsx
jorgemoya: I agree, will try to change this.

apps/web/vibes/soul/sections/product-detail/index.tsx
jorgemoya: Lets think about this. Purchase buttons can be disabled from the API if the product is not available, but the label of the button also depends on this status or if the order is preorderable. We want this to be streameable, since we fetch this state from the API. If this is the case, does having an object makes sense to you? Or should both props just be Streamable props?

apps/web/vibes/soul/sections/product-detail/index.tsx
jorgemoya: Fixed.

apps/web/vibes/soul/sections/product-detail/index.tsx
jorgemoya: I split into two props as suggested. I think this makes sense.

migueloller: > When a product is out of stock.

That makes sense. Thank you!

re: streaming. I think it's ok to have separate props. Something I want to encourage more in VIBES is combining streamable values when it makes sense:

```tsx
// In a server component
<Streamable value={Promise.all([streamableFoo, streamableBar]}>

// In a client component
<Streamable value={[streamableFoo, streamableBar]}>
```

The server component example is one I often do. The client component one I have to figure out if it's possible while keeping a reasonable API.

apps/web/vibes/soul/sections/product-detail/index.tsx
```diff
@@ -24,21 +24,23 @@ interface Props<F extends Field> {
   product: Streamable<ProductDetailProduct | null>;
   action: ProductDetailFormAction<F>;
   fields: Streamable<F[]>;
-  ctaLabel?: string;
   quantityLabel?: string;
   incrementLabel?: string;
   decrementLabel?: string;
+  submitButtonLabel?: Streamable<string>;
+  submitButtonDisabled?: Streamable<boolean>;
```
migueloller, line 31:
```suggestion
  submitButtonLabel?: Streamable<string | null>;
  submitButtonDisabled?: Streamable<boolean | null>;
```

apps/web/vibes/soul/sections/product-detail/index.tsx
```diff
@@ -97,12 +99,20 @@ export function ProductDetail<F extends Field>({
...
+                    value={Promise.all([
+                      streamableFields,
+                      streamableSubmitButtonLabel,
+                      streamableSubmitButtonDisabled,
+                    ])}
+                  >
+                    {([fields, submitButtonLabel, submitButtonDisabled]) => (
                       <ProductDetailForm
                         action={action}
-                        ctaLabel={ctaLabel}
```
migueloller, line 104: Seeing this diff makes me wonder, should we name the props `ctaLabel` and `ctaDisabled`? Or does "submit" make more sense in this context than "call to action"?

apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
```diff
@@ -53,6 +54,7 @@ export function ProductDetailForm<F extends Field>({
   quantityLabel = 'Quantity',
   incrementLabel = 'Increase quantity',
   decrementLabel = 'Decrease quantity',
+  disabled = false,
```
migueloller, line 57: Do we want to use this just for the CTA and name it `ctaDisabled`?

Also, we probably want some form of consistency between the prop names in this component and the section component.

apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
```diff
@@ -109,6 +111,7 @@ export function ProductDetailForm<F extends Field>({
             <NumberInput
               aria-label={quantityLabel}
               decrementLabel={decrementLabel}
+              disabled={disabled}
```
migueloller, line 114: Is it necessary to disable the quantity input as well? That seems reasonable but if we rename to `ctaDisabled` it might be confusing. If we keep the name `disabled`, then do we want to disable all fields in the form? Would that make things feel broken as opposed so just disabling the submit button?

apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
migueloller: I wonder if we should use the naming `addToCart` since that's really what this is designed for. Or are there other instances where perhaps you could immediately check out from a PDP? 🤔

migueloller (approved):
Got lots of comments here but this mostly looks good 👍🏻

apps/web/vibes/soul/sections/product-detail/index.tsx
jorgemoya: Either way is fine by me.

apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
jorgemoya: I will keep enabled for now and we can revisit.

migueloller (approved):

PR #397, opened 2024-12-13
feat: avoid layout shift when breadcrumbs are empty

hunterbecton:
## What/Why
- Avoids layout shift when no breadcrumbs are present by adding an empty `div`
- Removes padding from animate links to avoid extra height being added

## Testing

https://github.com/user-attachments/assets/0bcb17b5-3f4e-497c-ab63-07fdf0065361


migueloller (approved):
Nice!

migueloller (commented):
Actually, @hunterbecton, I believe the problem is not with the breadcrumbs, but with the PLP. We don't render breadcrumbs at all if we have less than 2 items. This means that this change won't help there, because we won't even render the component in that case.

This change is still useful in case an empty array is passed in but we still need to address the PLP problem.

PR #398, opened 2024-12-13
feat: add button css variables, add shape prop

apledger:


apps/web/app/(preview)/style/globals.css
```diff
@@ -5,6 +5,37 @@
...
+    --button-tertiary-border: hsl(var(--contrast-200));
+    --button-tertiary-focus: hsl(var(--foreground));
+
+    --button-ghost-background: transparent;
+    --button-ghost-background-hover: hsl(var(--foreground) / 5%);
+    --button-ghost-foreground: hsl(var(--foreground));
+    --button-ghost-foreground-hover: hsl(var(--foreground));
+    --button-ghost-border: transparent;
+    --button-ghost-focus: hsl(var(--foreground));
+  }
```
migueloller, line 38: Don't we want to apply this via the Soul schema and the CSS that's injected by that specific Soul brand?

migueloller (commented):
Changes to the component look good but don't we also want to update the VIBE's schema and configuration so that these values are part of it?

migueloller:
There seems to be a type issue that's causing CI to fail

apps/web/app/(preview)/style/globals.css
apledger: Nope because the Soul brand schemas should be just updating the underlying design tokens that power this theme!

migueloller (approved):

PR #399, opened 2024-12-14
fix: pass in server side default values for hydration issue

jorgemoya:
In ProductDetail, during server render, some of the Radix primitives get set as unchecked/unselected even when we know that these fields should have a value. During hydration, we get an error because aria attributes/params mismatch once values set.

This fix is to set a default value so that this mismatch is prevented. 

My testing included using non radix form elements which rendered fine, probably because properties don't change. Default values in the form get set correctly on static rendering but for dynamic rendering we don't see these for some reason.

Is it expected that the `useForm` hook ignores these values during server render?

I'm open to suggestions.

### DIFF
--- a/apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
+++ b/apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
@@ -102,6 +102,7 @@ export function ProductDetailForm<F extends Field>({
                 formField={formFields[field.name]!}
                 // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
                 key={formFields[field.name]!.id}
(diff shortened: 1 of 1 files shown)


migueloller (commented):
I think this is a fine approach but we need to make sure the default value we're passing matches the expected value. For example, checkboxes sometimes use `defaultChecked` or the value might be `'true'` when conform sets it to `'on'` for checkboxes.

As we discussed in Slack, the issue seems to be caused by using the edge runtime for this page in Catalyst. Perhaps we could address it by changing that instead?

jorgemoya:
Closing in favor of just removing `edge` runtime.

PR #401, opened 2024-12-15
Polish header search results and add some ESLint rules

migueloller:
- Sets search form submission pending state while submitting
- Handles errors when searching
- Adds ESLint rules for self-closing JSX tags, TypeScript array syntax, and sorting imports


apledger (approved):
LGTM!

PR #409, opened 2024-12-16
fix: aria-label and placeholder in Select

hunterbecton:
## What/Why
- Adds `aria-label` into Radix `SelectPrimitive.Trigger`
- Adds `aria-label` into Radix `SelectPrimitive.Value`
- Moves `placeholder` to `SelectPrimitive.Value`

Reference: [Radix Select Docs](https://www.radix-ui.com/primitives/docs/components/select#with-a-placeholder)

## Testing
Locally


apps/web/vibes/soul/form/select/index.tsx
```diff
@@ -51,7 +52,7 @@ export function Select({
           onBlur={onBlur}
           onFocus={onFocus}
         >
-          <SelectPrimitive.Value placeholder={placeholder} />
+          <SelectPrimitive.Value aria-label={value} placeholder={placeholder} />
```
migueloller, line 55: Feels weird to have the `aria-label` as the value here since the value might not be a user-readable string

jorgemoya: I agree, I don't think we need this one here, only the one in the `Trigger` component.

apps/web/vibes/soul/sections/products-list-section/sorting.tsx
```diff
@@ -58,6 +58,7 @@ function SortingInner({
```
jamesqquick, line 58: This wasn't changed in the PR, but do you know why we don't use a default value for `label` in the prop definition instead of giving the default here? So

migueloller: Because the value might be a promise that resolves to `null` so defaulting the prop when destructuring it would only handle the case when the prop isn't passed in, not the `null` case

PR #411, opened 2024-12-16
feat: add default props and optimistic update to account settings

chanceaclark:
### What/Why?

- Adds default values to be passed into the update account settings form.
- Adds a company field to the form
- Uses optimistic updates when submitting the update customer form action.
- Updates some examples to use the `AccountLayout` plus real links (to prevent duplicate React keys warning).

### Testing
Screenshot 2024-12-16 at 14 31 26

https://github.com/user-attachments/assets/10427720-3359-4096-9114-b993badeeba4

### DIFF
--- a/apps/web/vibes/soul/sections/account-settings-section/update-account-form.tsx
+++ b/apps/web/vibes/soul/sections/account-settings-section/update-account-form.tsx
@@ -2,35 +2,74 @@
 
 import { getFormProps, getInputProps, SubmissionResult, useForm } from '@conform-to/react';
 import { getZodConstraint, parseWithZod } from '@conform-to/zod';
-import { useActionState, useEffect } from 'react';
-import { useFormStatus } from 'react-dom';
+import { useActionState, useEffect, useOptimistic, useTransition } from 'react';
+import { z } from 'zod';
 
 import { Input } from '@/vibes/soul/form/input';
 import { Button } from '@/vibes/soul/primitives/button';
+import { toast } from '@/vibes/soul/primitives/toaster';
 
 import { updateAccountSchema } from './schema';
 
-type Action<State, Payload> = (state: Awaited<State>, payload: Payload) => State | Promise<State>;
+type Action<S, P> = (state: Awaited<S>, payload: P) => S | Promise<S>;
 
-export type UpdateAccountAction = Action<SubmissionResult | null, FormData>;
+export type UpdateAccountAction = Action<State, FormData>;
+
+export type Account = z.infer<typeof updateAccountSchema>;
+
+interface State {
+  account: Account;
+  successMessage?: string;
+  lastResult: SubmissionResult | null;
+}
 
 interface Props {
   action: UpdateAccountAction;
+  account: Account;
   firstNameLabel?: string;
   lastNameLabel?: string;
   emailLabel?: string;
+  companyLabel?: string;
   submitLabel?: string;
 }
 
 export function UpdateAccountForm({
   action,
+  account,
   firstNameLabel = 'First name',
   lastNameLabel = 'Last name',
   emailLabel = 'Email',
+  companyLabel = 'Company',
   submitLabel = 'Update',
 }: Props) {
-  const [lastResult, formAction] = useActionState(action, null);
+  const [state, formAction] = useActionState(action, { account, lastResult: null });
+  const [pending, startTransition] = useTransition();
+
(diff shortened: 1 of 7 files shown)


apps/web/vibes/soul/sections/account-settings-section/update-account-form.tsx
```diff
@@ -67,17 +148,15 @@ export function UpdateAccountForm({
...
-
-function SubmitButton({ children }: { children: React.ReactNode }) {
-  const { pending } = useFormStatus();
-
-  return (
-    <Button loading={pending} size="small" type="submit" variant="secondary">
-      {children}
-    </Button>
-  );
-}
```
chanceaclark, line 83: I'm curious if I should add this back in to get the form status state + button loading?

jorgemoya: I would say yes.

migueloller:
Nice work @chanceaclark!

Looking at the testing video I think something that's missing is a pending indicator while the request is in-flight. Something that I suggested we do for addresses that I don't know if Alan got to was adding a spinner next to the section title.

In this case it would involve adding a spinner next to "Account Settings".

apledger (approved):

migueloller (approved):
Looking good 👍🏻

apps/web/vibes/soul/sections/account-settings-section/update-account-form.tsx
```diff
@@ -67,17 +150,22 @@ export function UpdateAccountForm({
         key={fields.email.id}
         label={emailLabel}
       />
-      <SubmitButton>{submitLabel}</SubmitButton>
+      <Input
+        {...getInputProps(fields.company, { type: 'text' })}
+        errors={fields.company.errors}
+        key={fields.company.id}
+        label={companyLabel}
+      />
```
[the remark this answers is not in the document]
chanceaclark: I initially planned and changed this component to account for that, but realized that BigCommerce doesn't have custom account fields, only custom address and register customer fields. Because it wasn't dynamic and there are only a few fields here, I opted to keep it simple but if we see a need in the future we can probably shift it back over to dynamic fields.

apps/web/vibes/soul/sections/account-settings-section/update-account-form.tsx
chanceaclark: Let's tackle this separately as I think we need to rethink how we surface the hooks for alerting. I don't have a good idea for it yet, but if y'all have suggestions let me know.

apps/web/vibes/soul/sections/account-settings-section/update-account-form.tsx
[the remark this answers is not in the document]
migueloller: Yup, not something for us to worry about right now. VIBES has only 1 user, and it's Catalyst. If we have a need for different fields we can always use `DynamicForm` and have the consumer of VIBES pass in the fields, even if they're statically defined and not coming from an API

apps/web/vibes/soul/sections/account-settings-section/update-account-form.tsx
migueloller: For alerting, we do this right now:

```ts
useEffect(() => {
  if (form.status === 'success') toaster.success('Succeeded!')
}, [form.status])
```

chanceaclark: Ah... so that easier than I thought. For the localized strings, do we just pass it in as a prop to the component?

migueloller: For localization, do this:

```ts
const [{ lastResult, successMessage }, formAction] = useActionState(action, { lastResult: null })

const [form] = useForm({ lastResult })

useEffect(() => {
  if (lastResult.status === 'success') toaster.success(lastResult.successMessage)
}, [lastResult])
```

migueloller: We originally were passing it as a prop but we then realized that sometimes the localized message depends on the parameters. For example, in the PLP the empty state title uses the search term. We couldn't just pass it as a prop because the header is rendered in `layout.tsx`, which doesn't have access to `searchParams`. The solution was to have the message be part of the action state/payload. And this makes sense since the success state (or an error or empty state) are the outcome of the action

migueloller (approved):

jorgemoya (approved):

PR #415, opened 2024-12-17
feat: disable deep linking for non multi select inputs

jorgemoya:
+ Updated skeletons to prevent issues when importing to Catalyst
+ Add `FormStateInput` to hold state between reload
+ Disable deep linking for non multi select inputs that could fetch new variants.
+ Prevents entire form rerenders when typing on inputs.
+ Also fixes an issue with uncontrolled components when they default to empty string

## Before:
Screenshot 2024-12-17 at 11 48 26 AM

## After:
Screenshot 2024-12-17 at 11 50 13 AM

### DIFF
--- a/apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
+++ b/apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
@@ -11,7 +11,7 @@ import {
 } from '@conform-to/react';
 import { getZodConstraint, parseWithZod } from '@conform-to/zod';
 import { usePathname, useRouter } from 'next/navigation';
-import { createSerializer, parseAsString, useQueryState, useQueryStates } from 'nuqs';
+import { createSerializer, parseAsString, useQueryStates } from 'nuqs';
 import { ReactNode, useActionState, useCallback, useEffect } from 'react';
 import { useFormStatus } from 'react-dom';
 import { z } from 'zod';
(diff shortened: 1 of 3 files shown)


apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
```diff
@@ -39,7 +40,7 @@ interface State<F extends Field> {
 
 export type ProductDetailFormAction<F extends Field> = Action<State<F>, FormData>;
 
-interface Props<F extends Field> {
+interface Props<F extends Field & { deepLink?: boolean }> {
```
migueloller, line 43: We shouldn't need the generic here if we configure it in `FormField`

apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
```diff
@@ -175,10 +177,13 @@ function FormField({
 
   const handleChange = useCallback(
     (value: string) => {
-      void setParam(value);
+      if (field.deepLink === true) {
+        void setParam(value);
```
migueloller, line 181: Instead of conditionally calling `setParam` we probably want to completely exclude the parser from `nuqs`

apps/web/vibes/soul/sections/product-detail/schema.ts
```diff
@@ -5,6 +5,7 @@ interface FormField {
   label?: string;
   errors?: string[];
   required?: boolean;
+  deepLink?: boolean;
```
migueloller, line 8: Let's rename this to `persist`

apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
jorgemoya: This is not performant, instead of doing nothing, this actually triggers a rerender of the form as if `params` would've changed. 

I will stick with the conditional logic.

jorgemoya: Changed to use `useQueryStates` which works as expected.

migueloller (approved):

PR #416, opened 2024-12-17
fix: properly handle disabled state in NumberInput

jorgemoya:
`disabled` was undefined so the active/disabled classes were not being properly set.

Before:
Screenshot 2024-12-17 at 12 07 43 PM

After:
Screenshot 2024-12-17 at 12 07 50 PM

### DIFF
--- a/apps/web/vibes/soul/form/number-input/index.tsx
+++ b/apps/web/vibes/soul/form/number-input/index.tsx
@@ -17,7 +17,16 @@ export const NumberInput = React.forwardRef<
   }
 >(
   (
-    { label, className, required, errors, decrementLabel, incrementLabel, disabled, ...rest },
+    {
+      label,
+      className,
+      required,
+      errors,
+      decrementLabel,
+      incrementLabel,
+      disabled = false,
+      ...rest
+    },
     ref,
   ) => {
     const id = React.useId();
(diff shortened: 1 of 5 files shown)


jorgemoya:
Included in this PR the translation labels that were missing as well.

apps/web/vibes/soul/form/number-input/index.tsx
```diff
@@ -17,7 +17,16 @@ export const NumberInput = React.forwardRef<
...
   (
-    { label, className, required, errors, decrementLabel, incrementLabel, disabled, ...rest },
+    {
+      label,
+      className,
+      required,
+      errors,
+      decrementLabel,
+      incrementLabel,
+      disabled = false,
```
migueloller, line 27: This should be the only change necessary, no? We don't also need to change the conditional rendering logic to "defend" against `undefined`

migueloller (approved):

apps/web/vibes/soul/form/number-input/index.tsx
jorgemoya: Everything else was auto fixed by lint.

migueloller: LOL, pwned. Didn't know that was a thing 👍🏻

jorgemoya: 😆

fikrikarim (approved):

PR #418, opened 2024-12-17
feat: add breadcrumbs as prop to BlogPostContent, make 'author' optional

jordanarldt:
When working on implementing BlogPostContent into catalyst, I noticed that `breadcrumbs` were always static because it is not a prop on the component.

We should allow breadcrumbs as a prop so that they can be provided to the component.

### Summary
- Add `breadcrumbs` as prop
- Make `author` optional

### DIFF
--- a/apps/web/vibes/soul/sections/blog-post-content/index.tsx
+++ b/apps/web/vibes/soul/sections/blog-post-content/index.tsx
@@ -1,7 +1,7 @@
 import { clsx } from 'clsx';
 import Image from 'next/image';
 
-import { Breadcrumbs } from '@/vibes/soul/primitives/breadcrumbs';
+import { Breadcrumb, Breadcrumbs } from '@/vibes/soul/primitives/breadcrumbs';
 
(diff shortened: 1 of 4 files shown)


apledger (approved):
Looks good @jordanarldt! Approved.

apps/web/vibes/soul/sections/blog-post-content/index.tsx
```diff
@@ -36,33 +37,24 @@ export const BlogPostContent = function BlogPostContent({
...
             {title}
           </h1>
           <p>
-            {date} <span className="px-1">•</span> {author}
+            {date}{' '}
+            {Boolean(author) && (
+              <>
+                <span className="px-1">•</span> {author}
+              </>
+            )}
```
migueloller, line 57: We seem to be rendering a space regardless of whether the author is present or not. We should move that space inside the conditional. Also, instead of coercing using the `Boolean` constructor, can we check `auth != null` to be consistent?

PR #420, opened 2024-12-17
feat: use Stream inside Sorting and update props and focus-visible styles

hunterbecton:
## What / Why
- Updates `Sorting` component to use `Stream` instead of `Suspense`
- Adds `focus-visible` styles to `SelectPrimitive.Item`
- Updates imports to use `@/vibes`
- Adds `placeholder` prop into `ProductListSection`, `Sorting`, and `Select`
- Adds optional `hideLabel` prop to `Select`
- Passes `Promise.all` into stream values

This component encapsulates `Stream` inside of `Sorting` instead of using `Stream` inside the `ProductsListSection`. Is there a reason we don't do this for the other components like `Breadcrumbs`?

The component rendered inside the `Stream` is called `SortingResolved.` I think this makes the streamable pattern very clear to the user.

## Testing

https://github.com/user-attachments/assets/b333ba91-6c63-41ca-8625-7f9d5bafa30a

### DIFF
--- a/apps/web/vibes/soul/sections/products-list-section/sorting.tsx
+++ b/apps/web/vibes/soul/sections/products-list-section/sorting.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import { parseAsString, useQueryState } from 'nuqs';
-import { Suspense, use, useOptimistic } from 'react';
+import { use, useOptimistic } from 'react';
 
 import { Select } from '@/vibes/soul/form/select';
 import { Streamable, useStreamable } from '@/vibes/soul/lib/streamable';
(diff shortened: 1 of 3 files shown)


jorgemoya (approved):
Ideally we should have two different commits/prs, one for Select and one for Sorting, but this looks good to me!

apps/web/vibes/soul/sections/products-list-section/sorting.tsx
```diff
@@ -14,50 +14,62 @@ export interface Option {
...
-      <SortingInner
-        defaultValue={defaultValue}
-        label={label}
-        options={options}
-        paramName={paramName}
-      />
-    </Suspense>
+    <Stream
+      fallback={<SortingSkeleton />}
+      value={Promise.all([streamableLabel, streamableOptions, streamablePlaceholder])}
```
migueloller, line 32: We cannot use `Promise.all` in a client component because that will make a new promise on render which will cause React to throw an error. Did this not throw when testing locally with the examples?

migueloller (changes requested):
Looks good except for creating a new promise in a client component. I'm surprised you didn't run into issues when testing locally

apps/web/vibes/soul/sections/products-list-section/sorting.tsx
hunterbecton: No errors on my end, but I moved the the `Stream` component inside into `ProductListSection` so I could use `Promise.all()`.

migueloller (approved):
LGTM! 👍🏻

PR #421, opened 2024-12-18
feat: add CSS variables and props to Navigation

apledger:
In this PR I've added CSS variables for every color value used in the `Navigation` component except for the ones used by fallback skeletons.

We need to make a decision on whether we're going to allow theming for placeholder skeletons or if we should use some mix of color difference and alpha, or control all skeletons items globally. Either way, this can be solved later.

I've also added some props to the Navigation component per @andrewreifman's request:
- `logoWidth`
- `mobileLogo`
- `mobileLogoWidth`
- `linksPosition`

These have been thoroughly tested locally. I'm not going to record me testing every single variable but know that I did.

https://github.com/user-attachments/assets/e0f00d04-e256-4bc6-b7d1-bb85176e5b21


fikrikarim (approved):
nice

PR #425, opened 2024-12-18
feat: use streamable props on BlogPostList

jordanarldt:
### Summary
- Ensure `posts` property is `Streamable<BlogPost[]>` on FeaturedBlogPostList and BlogPostList
- Update all previews with a timeout to ensure the loading state is displayed in the docs
- Add `object-cover` style to BlogPostCard to prevent image stretching
- Minor refactoring

### DIFF
--- a/apps/web/vibes/soul/sections/featured-blog-post-list/index.tsx
+++ b/apps/web/vibes/soul/sections/featured-blog-post-list/index.tsx
@@ -1,3 +1,4 @@
+import { Streamable } from '@/vibes/soul/lib/streamable';
 import { BlogPost } from '@/vibes/soul/primitives/blog-post-card';
 import { ButtonLink } from '@/vibes/soul/primitives/button-link';
 import { BlogPostList } from '@/vibes/soul/sections/blog-post-list';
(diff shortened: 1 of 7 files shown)


apps/web/vibes/soul/sections/featured-blog-post-list/index.tsx
```diff
@@ -11,15 +12,10 @@ interface Props {
...
+  posts: Streamable<BlogPost[]>;
 }
 
-export const FeaturedBlogPostList = function FeaturedBlogPostList({
-  title,
-  description,
-  cta,
-  posts,
-}: Props) {
+export function FeaturedBlogPostList({ title, description, cta, posts: streamablePosts }: Props) {
```
migueloller, line 18: 🍹 no need to alias to `streamablePosts` since we pass it straight down to `BlogPostList`. The reason we usually alias it is to work around the `no-shadow` ESLint rule when it complains we name the streamed result the same as the prop (i.e., having a prop `posts`, but then also naming the streamed value in the `Streamable` callback `posts`)

migueloller (approved):
LGTM 👍🏻 

Looks like some other stylistic changes sneaked in in there. Thanks!

apps/web/vibes/soul/sections/featured-blog-post-list/index.tsx
jordanarldt: @migueloller Ahh, makes sense! I'll update this real quick

apledger (approved):
Great job! LGTM

PR #426, opened 2024-12-18
feat: address-list-section updates

chanceaclark:
## What/why?

> [!TIP]
> Hide whitespace might help reviewing.

Adds two bits of functionality to the `AddressListSection` component:
- Makes setting the default address optional as some systems may not have that ability.
- Adds a `minimumAddressCount` prop to enforce how many addresses can be deleted.

### Set Default Logic
Below is the matrix on the prop logic:
| `defaultAddress` value | Functionality |
| ----------------------- | ------------- |
| `undefined`                   | No default address functionality will be shown |
| `{ id: null }`                    | Shows the `Set Default` button in a situation where a customer doesn't have a default address set |
| `{ id: '1' }`                     | Renders the `Set Default` button on every address that is not the default. Shows the default label on the default address. |

### Set Default Logic:
Screenshot 2024-12-18 at 15 26 57
Screenshot 2024-12-18 at 15 27 10
Screenshot 2024-12-18 at 15 27 23

### Minimum Address Count:

https://github.com/user-attachments/assets/2fa80ce7-5dc2-4a4a-b64c-9a2c9ac76856

### DIFF
--- a/apps/web/vibes/soul/examples/sections/address-list-section/actions.ts
+++ b/apps/web/vibes/soul/examples/sections/address-list-section/actions.ts
@@ -1,23 +1,21 @@
 import { SubmissionResult } from '@conform-to/react';
 import { parseWithZod } from '@conform-to/zod';
 import { randomUUID } from 'crypto';
-import { z } from 'zod';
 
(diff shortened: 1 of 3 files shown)


jorgemoya (approved):
This makes sense to me. Looking good!

PR #429, opened 2024-12-19
Fix image priority on PLP to improve lighthouse

bookernath:
Mirroring https://github.com/bigcommerce/catalyst/pull/1781

## What/Why?
Improve lighthouse finding around above-the-fold image being loaded lazily; prioritizing just the first image is enough to eliminate the finding and maps to the mobile experience where you can only see one image. On desktop this might ideally be more like the first 3-4 images, but the finding on desktop goes away also so I'm fine to start with 1 image.

## Testing
Before:

image

After (finding gone):

image

### DIFF
--- a/apps/web/vibes/soul/primitives/products-list/index.tsx
+++ b/apps/web/vibes/soul/primitives/products-list/index.tsx
@@ -60,7 +60,7 @@ export function ProductsList({
           return (
             <div className={clsx('w-full @container', className)}>
               <div className="mx-auto grid grid-cols-1 gap-x-4 gap-y-6 @sm:grid-cols-2 @2xl:grid-cols-3 @2xl:gap-x-5 @2xl:gap-y-8 @5xl:grid-cols-4 @7xl:grid-cols-5">
-                {products.map((product) => (
+                {products.map((product, index) => (
                   <ProductCard
                     aspectRatio={aspectRatio}
                     colorScheme={colorScheme}
@@ -69,6 +69,7 @@ export function ProductsList({
                     key={product.id}
                     product={product}
                     showCompare={showCompare}
+                    imagePriority={index === 0}
                   />
                 ))}
               </div>


apps/web/vibes/soul/primitives/products-list/index.tsx
```diff
@@ -69,6 +69,7 @@ export function ProductsList({
                     key={product.id}
                     product={product}
                     showCompare={showCompare}
+                    imagePriority={index === 0}
```
migueloller, line 72: What is instead we made it so that `ListProduct` was `CardProduct & { imagePriority?: boolean }` and then from Catalyst we could pick exactly which images we'd like to prioritize?

bookernath: Yep I had similar thoughts about the layering... counterargument is that this change within the primitive helps lighthouse for anyone that uses these components.

I think your argument is overall better

migueloller: The primitive might be rendered below the fold, in which case we don't want the first image to have priority. Keep in mind that this primitive might be used in places other than the PLP

migueloller: So what do you think @bookernath? Are you down to try my suggestion or did you want to move forward with the changes you have right now?

jorgemoya: I prefer if we keep the primitive agnostic of this logic and just pass default from section or, like @migueloller suggested, optional config that can be passed from Catalyst.

migueloller: And just to be more explicit. My suggested changes would involve changing this line: https://github.com/makeswift/vibes/pull/429/files/a8b79ac5fcc35fa1f32eefc8892f7617782ae09d#diff-b96599c391d195af1a5bd0759ff7add174153291e25192fd05136b0505bf31d1R12

To this:

```ts
export type ListProduct = CardProduct & { imagePriority?: boolean };
```

Then, from Catalyst we can pass the products as follows:

```tsx
<ProductsListSection products={products.map((product, index) => ({ ...product, imagePriority: index <= 3 })} />
```

PR #430, opened 2024-12-19
feat: refactor CursorPagination, add breadcrumbs to FeaturedBlogPostList

jordanarldt:
## Summary
- Refactored `CursorPagination` component and it's uses
- Added `breadcrumbs` to FeaturedBlogPostList

## Testing


apps/web/vibes/soul/primitives/cursor-pagination/index.tsx
```diff
@@ -26,9 +25,9 @@ interface Props {
 
 export function CursorPagination(props: Props) {
   return (
-    <Suspense fallback={<CursorPaginationSkeleton />}>
-      <CursorPaginationResolved {...props} />
-    </Suspense>
+    <Stream fallback={<CursorPaginationSkeleton />} value={props}>
+      {CursorPaginationResolved}
+    </Stream>
```
migueloller, line 30: This isn't the proper use of `Stream` as it's calling `CursorPaginationResolved` instead of rendering it as a component, which would break the rules of hooks, have issues with RSC/client component boundaries, etc. Also, `value` should be a streamable value, not all props.

Instead, we should use leave this unchanged as the `useStreamable` calls in `CursorPaginationResolved` will suspend, triggering the Suspense boundary and fallback.

apps/web/vibes/soul/sections/featured-blog-post-list/index.tsx
```diff
@@ -1,39 +1,48 @@
...
-        <h1 className="mb-3 font-heading text-4xl font-medium leading-none text-foreground @xl:text-5xl @4xl:text-6xl">
-          {title}
-        </h1>
-
-        {description != null && description !== '' && (
-          <p className="max-w-lg text-lg text-contrast-500">{description}</p>
+        {streamableBreadcrumbs && (
+          <Stream fallback={<BreadcrumbsSkeleton />} value={streamableBreadcrumbs}>
+            {(breadcrumbs) =>
+              Boolean(breadcrumbs.length) && <Breadcrumbs breadcrumbs={breadcrumbs} />
```
migueloller, line 28: The breadcrumbs already handle a length of zero so we shouldn't have to check that here

migueloller (changes requested):

apps/web/vibes/soul/sections/featured-blog-post-list/index.tsx
jordanarldt: Ahh good point

migueloller (approved):

apps/web/vibes/soul/primitives/breadcrumbs/index.tsx
```diff
@@ -9,11 +10,23 @@ export interface Breadcrumb {
...
+  breadcrumbs: Streamable<Breadcrumb[]>;
   className?: string;
 }
 
-export function Breadcrumbs({ breadcrumbs, className }: BreadcrumbsProps) {
+export function Breadcrumbs(props: BreadcrumbsProps) {
+  return (
+    <Stream fallback={<BreadcrumbsSkeleton />} value={props.breadcrumbs}>
+      {(breadcrumbs) => (
+        <BreadcrumbsResolved breadcrumbs={breadcrumbs} className={props.className} />
```
jorgemoya, line 21: I, personally, would remove BreadcrumbsResolved and just copy and paste code inside this Stream component.

jordanarldt:
> Minor nitpick and probably an annoyance 😅 , but I would prefer if we split multiple changes to various PRs, or at the very least, have a single commit for each change. For now this is okay, but just a suggestion for the future!

For sure! Yeah I usually do that, but I got a bit carried away knocking out several tasks at once 😆

apps/web/vibes/soul/sections/products-list-section/index.tsx
```diff
@@ -66,9 +62,7 @@ export function ProductsListSection({
         <div className="mx-auto max-w-screen-2xl px-4 py-10 @xl:px-6 @xl:py-14 @4xl:px-8 @4xl:py-12">
           <div>
             <Stream fallback={<BreadcrumbsSkeleton />} value={streamableBreadcrumbs}>
```
jorgemoya, line 64: Isn't this `Stream` no longer needed? Just call `<Breadcrumbs />` directly.

jorgemoya (commented):
Looking good, thanks for looking into this!

apps/web/vibes/soul/sections/products-list-section/index.tsx
jordanarldt: @jorgemoya The `ProductListSection` expects `breadcrumbs` to be `breadcrumbs?: Streamable<Breadcrumb[] | null>;`

I can fix this though

apps/web/vibes/soul/sections/products-list-section/index.tsx
```diff
@@ -65,11 +61,8 @@ export function ProductsListSection({
       <div className="@container">
         <div className="mx-auto max-w-screen-2xl px-4 py-10 @xl:px-6 @xl:py-14 @4xl:px-8 @4xl:py-12">
           <div>
-            <Stream fallback={<BreadcrumbsSkeleton />} value={streamableBreadcrumbs}>
-              {(breadcrumbs) =>
-                breadcrumbs && breadcrumbs.length > 1 && <Breadcrumbs breadcrumbs={breadcrumbs} />
-              }
-            </Stream>
+            {breadcrumbs && <Breadcrumbs breadcrumbs={breadcrumbs} />}
```
migueloller, line 64: For this we _do_ want to keep this logic since it's avoiding to render a breadcrumbs of length 1 just for this section. We could make this configurable in a different way, e.g., a prop on `Breadcrumbs` to control when we show the empty state (basically configuring the `length === 0` check). Regardless, there seems to be behavior being changed here

migueloller (approved):

PR #433, opened 2024-12-20
feat: add summary and accordions to ProductDetails

jorgemoya:
Update designs of product details to include accordions and summary.

https://github.com/user-attachments/assets/5a11a598-182f-44cd-88cc-5f746a748177

https://github.com/user-attachments/assets/bfadd177-4359-48ea-8173-da8bbc247f83


migueloller (approved):
This looks great! What do you think @jorgemoya, do you like this better than what we had before?

jorgemoya:
> This looks great! What do you think @jorgemoya, do you like this better than what we had before?

A lot better experience imo! Thanks for working on this with @andrewreifman.

PR #434, opened 2024-12-20
feat: migrate address form to use dynamic form

chanceaclark:
## What/Why?

Draft PR for migrating the AddressListSection over to using the `DynamicForm`.

### DIFF
--- a/apps/web/vibes/soul/primitives/dynamic-form/index.tsx
+++ b/apps/web/vibes/soul/primitives/dynamic-form/index.tsx
@@ -27,7 +27,7 @@ import { RadioGroup } from '@/vibes/soul/form/radio-group';
 import { Select } from '@/vibes/soul/form/select';
 import { SwatchRadioGroup } from '@/vibes/soul/form/swatch-radio-group';
 import { Textarea } from '@/vibes/soul/form/textarea';
-import { Button } from '@/vibes/soul/primitives/button';
+import { Button, Props as ButtonProps } from '@/vibes/soul/primitives/button';
 
 import { Field, FieldGroup, schema } from './schema';
 
@@ -43,17 +43,23 @@ export type DynamicFormAction<F extends Field> = Action<State<F>, FormData>;
 interface Props<F extends Field> {
   fields: Array<F | FieldGroup<F>>;
   action: DynamicFormAction<F>;
+  buttonSize?: ButtonProps['size'];
+  cancelLabel?: string;
   submitLabel?: string;
   submitName?: string;
   submitValue?: string;
+  onCancel?: (e: React.MouseEvent<HTMLButtonElement>) => void;
 }
 
 export function DynamicForm<F extends Field>({
   action,
   fields: defaultFields,
+  buttonSize = 'medium',
+  cancelLabel = 'Cancel',
   submitLabel = 'Submit',
   submitName,
   submitValue,
+  onCancel,
 }: Props<F>) {
   const [{ lastResult, fields }, formAction] = useActionState(action, {
     fields: defaultFields,
@@ -118,8 +124,18 @@ export function DynamicForm<F extends Field>({
 
             return <DynamicFormField field={field} formField={formField} key={formField.id} />;
           })}
-          <div className="flex gap-x-3 pt-3">
-            <SubmitButton name={submitName} value={submitValue}>
+          <div className="flex gap-1 pt-3">
+            {onCancel && (
+              <Button
+                aria-label={`${cancelLabel} ${submitLabel}`}
+                onClick={onCancel}
+                size={buttonSize}
+                variant="tertiary"
+              >
+                {cancelLabel}
+              </Button>
+            )}
+            <SubmitButton name={submitName} size={buttonSize} value={submitValue}>
               {submitLabel}
             </SubmitButton>
           </div>
@@ -138,22 +154,17 @@ function SubmitButton({
   children,
   name,
   value,
+  size,
 }: {
   children: React.ReactNode;
   name?: string;
   value?: string;
+  size: ButtonProps['size'];
 }) {
   const { pending } = useFormStatus();
 
   return (
-    <Button
-      className="w-auto @xl:w-56"
-      loading={pending}
-      name={name}
-      size="medium"
-      type="submit"
-      value={value}
-    >
+    <Button loading={pending} name={name} size={size} type="submit" value={value}>
       {children}
     </Button>
   );
--- a/apps/web/vibes/soul/primitives/dynamic-form/schema.ts
+++ b/apps/web/vibes/soul/primitives/dynamic-form/schema.ts
@@ -5,6 +5,7 @@ interface FormField {
   label?: string;
   errors?: string[];
   required?: boolean;
+  id?: string;
 }
 
 type RadioField = {
@@ -114,6 +115,11 @@ type ConfirmPasswordField = {
   type: 'confirm-password';
 } & FormField;
 
+type HiddenInputField = {
+  type: 'hidden';
+  defaultValue?: string;
+} & FormField;
+
 export type Field =
   | RadioField
   | CheckboxField
@@ -128,7 +134,8 @@ export type Field =
   | SelectField
   | PasswordField
   | ConfirmPasswordField
-  | EmailInputField;
+  | EmailInputField
+  | HiddenInputField;
 
 export type FieldGroup<F> = F[];
 
--- a/apps/web/vibes/soul/sections/address-list-section/index.tsx
+++ b/apps/web/vibes/soul/sections/address-list-section/index.tsx
@@ -6,10 +6,12 @@ import { startTransition, useActionState, useEffect, useOptimistic, useState } f
 import { useFormStatus } from 'react-dom';
 import { z } from 'zod';
 
-import { Input } from '@/vibes/soul/form/input';
 import { Badge } from '@/vibes/soul/primitives/badge';
 import { Button } from '@/vibes/soul/primitives/button';
+import { DynamicForm } from '@/vibes/soul/primitives/dynamic-form';
+import { Field, FieldGroup } from '@/vibes/soul/primitives/dynamic-form/schema';
 import { Spinner } from '@/vibes/soul/primitives/spinner';
+import { toast } from '@/vibes/soul/primitives/toaster';
 
 import { schema } from './schema';
 
@@ -19,20 +21,22 @@ export interface DefaultAddressConfiguration {
   id: string | null;
 }
 
-type Action<State, Payload> = (state: Awaited<State>, payload: Payload) => State | Promise<State>;
+type Action<S, P> = (state: Awaited<S>, payload: P) => S | Promise<S>;
 
-interface State<A extends Address> {
+interface State<A extends Address, F extends Field> {
   addresses: A[];
   defaultAddress?: DefaultAddressConfiguration;
   lastResult: SubmissionResult | null;
+  fields: Array<F | FieldGroup<F>>;
 }
 
-interface Props<A extends Address> {
+interface Props<A extends Address, F extends Field> {
   title?: string;
   addresses: A[];
+  fields: Array<F | FieldGroup<F>>;
   minimumAddressCount?: number;
   defaultAddress?: DefaultAddressConfiguration;
-  addressAction: Action<State<A>, FormData>;
+  addressAction: Action<State<A, F>, FormData>;
   editLabel?: string;
   deleteLabel?: string;
   updateLabel?: string;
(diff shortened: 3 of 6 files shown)


apps/web/vibes/soul/sections/address-list-section/index.tsx
```diff
@@ -141,132 +127,146 @@ export function AddressListSection<A extends Address>({
...
-                intent="update"
-                lastNameLabel={lastNameLabel}
-                onCancel={() =>
-                  setActiveAddressIds((prev) => prev.filter((id) => id !== address.id))
-                }
-                onSubmit={(formData) => {
-                  setActiveAddressIds((prev) => prev.filter((id) => id !== address.id));
+            <div className="w-[480px] space-y-4">
+              <DynamicForm
+                action={(_prevState, formData) => {
```
migueloller, line 132: For progressive enhancement, we should use the form action directly or from useActionState and use onSubmit for optimistic logic

apps/web/vibes/soul/sections/address-list-section/index.tsx
```diff
@@ -141,132 +127,146 @@ export function AddressListSection<A extends Address>({
...
-                    >
-                      {deleteLabel}
-                    </AddressActionButton>
-                  )}
 
-                  {optimisticState.defaultAddress &&
-                    optimisticState.defaultAddress.id !== address.id && (
+                  return field;
+                })}
+                onCancel={() => setShowNewAddressForm(false)}
```
migueloller, line 157: We should look into using conform's "intent" feature so that we don't need to use a callback that will force this to be a client component. We can still get instant behavior using optimistic state

migueloller (commented):
This is looking good! Just have a couple of comments

apps/web/vibes/soul/sections/address-list-section/index.tsx
chanceaclark: I don't think we will be able to join the visibility logic on the form with the form submission here (if we are trying to use one state for everything). The submission logic requires the form submission value to adhere the schema, which on creating a new address is completely empty. I think we need to keep the showing/hiding of the multiple dynamic forms separate (and simple) here.

Also, does it matter if it's a client component? A form using a server action has to be a client component anyways.

apps/web/vibes/soul/sections/address-list-section/index.tsx
chanceaclark: I believe that should be accounted for in the `DynamicForm` component: https://github.com/makeswift/vibes/blob/main/apps/web/vibes/soul/primitives/dynamic-form/index.tsx#L81-L86

apps/web/vibes/soul/primitives/dynamic-form/schema.ts
```diff
@@ -108,10 +106,17 @@ type ButtonRadioField = {
 
 type PasswordField = {
   type: 'password';
+  defaultValue?: string;
```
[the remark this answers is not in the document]
chanceaclark: It does feel weird here, but technically they can have default values. Initially I added this to prevent having to do some complex type checking, but looks like I can remove this.

apps/web/vibes/soul/primitives/dynamic-form/schema.ts
```diff
@@ -36,8 +36,6 @@ type NumberInputField = {
   min?: number;
   max?: number;
   step?: number;
-  incrementLabel?: string;
```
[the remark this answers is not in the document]
chanceaclark: I was getting some typing errors previously, seems like they were fixed somehow 🤔 I'll add them back.

apps/web/vibes/soul/sections/address-list-section/index.tsx
migueloller: The schema can have different payloads based on the intent. For example, the intent to change an address can require certain fields while the one to add a new address field or remove one can take no payload.

re: it being a client component, yeah, it's not a huge deal, it's more of a matter of consistency where there's various sections in VIBES we want to server components. In this case, since it's a form, it has to be a client component anyway to use `useActionState`.

jorgemoya (dismissed):

apps/web/vibes/soul/sections/address-list-section/index.tsx
```diff
@@ -127,6 +113,17 @@ export function AddressListSection<A extends Address>({
...
+    lastResult: state.lastResult,
+  });
+
+  useEffect(() => {
+    if (form.errors) {
+      form.errors.forEach((error) => {
+        toast.error(error);
+      });
+    }
+  }, [form.errors]);
```
chanceaclark, line 126: Added this to show any server errors for now.

chanceaclark (commented):
♻️ 🙏

apps/web/vibes/soul/primitives/dynamic-form/index.tsx
```diff
@@ -118,8 +124,18 @@ export function DynamicForm<F extends Field>({
...
+            {onCancel && (
+              <Button
+                aria-label={`${cancelLabel} ${submitLabel}`}
+                onClick={onCancel}
+                size={buttonSize}
+                variant="tertiary"
+              >
+                {cancelLabel}
+              </Button>
+            )}
```
migueloller, line 137: It feels weird to add a cancel button to the dynamic form primitive as opposed to adding it to the component that renders this form

apps/web/vibes/soul/primitives/dynamic-form/schema.ts
```diff
@@ -5,6 +5,7 @@ interface FormField {
   label?: string;
   errors?: string[];
   required?: boolean;
+  id?: string;
```
migueloller, line 8: Is it necessary to add this here vs it being handled by the generics?

apps/web/vibes/soul/sections/address-list-section/index.tsx
migueloller: If we prop we pass in isn't a server action, though, it won't work. Instead, `DynamicForm` should have an `onSubmit` prop IMO

apps/web/vibes/soul/sections/address-list-section/index.tsx
```diff
@@ -321,6 +321,9 @@ function AddressActionButton({
   onSubmit: (formData: FormData) => void;
 } & Omit<React.ComponentProps<'button'>, 'onSubmit'>) {
   const [form, fields] = useForm({
+    // @ts-expect-error The form requires index signature values to be of
+    // type 'string', 'null', or 'undefined' but the zod .passthrough() method
+    // returns the value 'unknown' for any index signature values.
```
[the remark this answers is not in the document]
migueloller: Just to make sure I understand, is the problem that we need to use `passthrough` to allow generic fields but that conflicts with `useForm`'s types?

apps/web/vibes/soul/sections/address-list-section/index.tsx
```diff
@@ -342,12 +357,18 @@ function AddressActionButton({
...
-      <input {...getInputProps(fields.state, { type: 'hidden' })} key={fields.state.id} />
+      <input
+        {...getInputProps(fields.stateOrProvince, { type: 'hidden' })}
+        key={fields.stateOrProvince.id}
+      />
       <input {...getInputProps(fields.postalCode, { type: 'hidden' })} key={fields.postalCode.id} />
-      <input {...getInputProps(fields.country, { type: 'hidden' })} key={fields.country.id} />
+      <input
+        {...getInputProps(fields.countryCode, { type: 'hidden' })}
+        key={fields.countryCode.id}
```
migueloller, line 370: What do we need all these hidden inputs for? Aren't there actual inputs that have these values?

apps/web/vibes/soul/sections/address-list-section/schema.ts
```diff
@@ -1,15 +1,17 @@
...
-  phone: z.string().optional(),
-  country: z.string(),
-});
+export const schema = z
+  .object({
+    id: z.string(),
+    firstName: z.string(),
+    lastName: z.string(),
+    company: z.string().optional(),
+    address1: z.string(),
```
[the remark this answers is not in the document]
migueloller: Can you expand more what you mean by "made consumer logic more simple"? Do you mean that in Catalyst we didn't have to map address fields?

migueloller (approved):

apps/web/vibes/soul/sections/address-list-section/index.tsx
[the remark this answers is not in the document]
chanceaclark: Correct

apps/web/vibes/soul/sections/address-list-section/index.tsx
chanceaclark: For the delete and set default buttons, we don't have a form, but we want to update the optimistic state here.

apps/web/vibes/soul/sections/address-list-section/schema.ts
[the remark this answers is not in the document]
chanceaclark: Yeah, this might be a problem for other ecom consumers of vibes. We had to map when we fetch the data and pass it into VIBES and remap it in server actions. It made the code a bit messy in Catalyst and I want to keep our reference implementation cleaner, hence renaming these.

apps/web/vibes/soul/primitives/dynamic-form/schema.ts
chanceaclark: I'm not sure what you mean here?

apps/web/vibes/soul/primitives/dynamic-form/index.tsx
chanceaclark: What's the pattern here then? Pass it in as a prop (`{ cancelButton: ReactNode }` or via `children`?

migueloller: I need to do some more thinking to form an opinion here. As it is right now it feels like we're coupling the needs of the address form with the dynamic form. I think it's completely fine given the time pressures we're feeling right now but later on I would like to rethink how we handle dynamic form fields. Perhaps the component doesn't even attempt to render a full form and just renders the fields, or perhaps the submit button (and cancel button) can be part of the dynamic data. Not sure just yet, so let's just punt it?

apps/web/vibes/soul/primitives/dynamic-form/schema.ts
migueloller: We've been using generics in the props so that data can be passed to the component and back to the server via server actions and I was thinking that this could be an extra field like that. Perhaps I'm misunderstanding, though, and this `id` field is used by VIBES? It being optional jumped out at me as something that might not be needed

apps/web/vibes/soul/sections/address-list-section/index.tsx
migueloller: I see. Feels a bit weird to have to do this but I don't have any solutions to offer right now so carry on 😅

apps/web/vibes/soul/primitives/dynamic-form/index.tsx
chanceaclark: I see what you mean there. I also see we could use `children` to allow consumers to define the layout of the buttons – the size and width of the buttons should be different between the address and register customer forms. But yeah, let's punt this decision.

apps/web/vibes/soul/primitives/dynamic-form/schema.ts
chanceaclark: So this was more for the consumers needing it. In our consumption, we were conflating `name` with the entity ID, and this doesn't work well when trying to build payloads for the API using `fields` on the state (specifically BigCommerce Custom Form Fields). This separates trying to identify form fields from the `name` to the `id` here.

chanceaclark:
Just going to throw up a tl'dr of what we want to do here in the future, but are not show-stoppers for merging this in:

- Separate out calling `onSubmit` and `action` in the `DynamicForm` component for a better pattern of progressive enhancement.
- Separate concerns for rendering buttons, e.g. submit or cancel, in the `DynamicForm` component. This way the consumers can handle some of the complexities (i18n, intent, client-side callbacks, etc...).
- Add a loading indicator next to the title on the `AddressListSection` component for when optimistic updates are happening.

Another piece we need to dive more on/consider for this component, is if optimistic should be used for adding/updating addresses for when API validation fails. We sort of solve this is a toast but should think about that user interaction more.

jorgemoya (approved):

apps/web/vibes/soul/primitives/dynamic-form/schema.ts
migueloller: I see. I think that we might be able to accomplish that by allowing extra fields if the fields are generic. I don't want to continue dragging this along, though, so again, this is something we can refine later

chanceaclark: `placeholder` is another attribute example of something that would be nice to add onto fields here.

migueloller: Agreed!

PR #436, opened 2024-12-23
feat: add Skeleton component and SkeletonWrapper component

hunterbecton:
## What/Why
This PR introduces a `Skeleton` primitive to VIBES that will help streamline the creation of skeleton loaders across the project. I'm also proposing a `Skeleton.Root` component to simplify creating component skeletons because it automatically handles accessibility and the `data-pending` pattern to handle skeleton animations.

This PR also promotes the `ProductsCarousel` to a section after discussing with the team that primitives should only be responsible for rendering data and that primitives should not compose other primitives.

### Skeleton component
- Provides `Skelton.Root` `Skeleton.Text` `Skelton.Box` and `Skeleton.Icon` components
- Simplifies the process of making skeletons accessible
- Relies on Tailwind classes to style the component rather than props that control the styles
- `characterCount` prop on `Skeleton.Text` to simplify how we're using `ch` to style
- Changed default skeleton color so it looks better on a larger range of backgrounds 

## Testing

https://github.com/user-attachments/assets/a656858d-d741-46bd-aae4-a10afa09178e

### DIFF
--- a/apps/web/vibes/soul/primitives/card/index.tsx
+++ b/apps/web/vibes/soul/primitives/card/index.tsx
@@ -3,6 +3,8 @@ import { ArrowUpRight } from 'lucide-react';
 import Image from 'next/image';
 import Link from 'next/link';
 
+import * as Skeleton from '@/vibes/soul/primitives/skeleton';
+
 export interface CardProps {
   className?: string;
   title: string;
@@ -123,9 +125,9 @@ export function CardSkeleton({
 }) {
   return (
     <div className={clsx('@container', className)}>
-      <div
+      <Skeleton.Box
         className={clsx(
-          'rounded-[var(--card-border-radius,1rem)] bg-contrast-100',
+          'rounded-[var(--card-border-radius,1rem)]',
           {
             '5:6': 'aspect-[5/6]',
             '3:4': 'aspect-[3/4]',
@@ -133,10 +135,8 @@ export function CardSkeleton({
           }[aspectRatio],
         )}
       />
-      <div className="mt-3 text-lg">
-        <div className="flex h-[1lh] items-center">
-          <span className="block h-[1ex] w-[10ch] rounded-sm bg-contrast-100" />
-        </div>
+      <div className="mt-3">
+        <Skeleton.Text className="rounded text-lg" characterCount={10} />
       </div>
     </div>
   );
--- a/apps/web/vibes/soul/primitives/product-card/index.tsx
+++ b/apps/web/vibes/soul/primitives/product-card/index.tsx
@@ -4,6 +4,7 @@ import Link from 'next/link';
 
 import { Badge } from '@/vibes/soul/primitives/badge';
 import { Price, PriceLabel } from '@/vibes/soul/primitives/price-label';
+import * as Skeleton from '@/vibes/soul/primitives/skeleton';
 
 import { Compare } from './compare';
 
@@ -172,9 +173,9 @@ export function ProductCardSkeleton({
 }) {
   return (
     <div className={clsx('@container', className)}>
-      <div
+      <Skeleton.Box
         className={clsx(
-          'flex flex-col gap-2 rounded-xl bg-contrast-100 @md:rounded-2xl',
+          'rounded-xl @md:rounded-2xl',
           {
             '5:6': 'aspect-[5/6]',
             '3:4': 'aspect-[3/4]',
@@ -183,18 +184,10 @@ export function ProductCardSkeleton({
         )}
       />
       <div className="mt-2 flex flex-col items-start gap-x-4 gap-y-3 px-1 @xs:mt-3 @2xl:flex-row">
-        <div className="flex-1">
-          <div className="flex flex-col text-base">
-            <div className="flex h-[1lh] items-center">
-              <span className="block h-[1ex] w-[10ch] rounded-sm bg-contrast-100" />
-            </div>
-            <div className="mb-2 flex h-[1lh] items-center text-sm font-normal text-contrast-400">
-              <span className="block h-[1ex] w-[8ch] rounded-sm bg-contrast-100" />
-            </div>
-            <div className="flex h-[1lh] items-center">
-              <span className="block h-[1ex] w-[5ch] rounded-sm bg-contrast-100" />
-            </div>
-          </div>
+        <div className="w-full text-base">
+          <Skeleton.Text className="rounded" characterCount={10} />
+          <Skeleton.Text className="rounded" characterCount={8} />
+          <Skeleton.Text className="rounded" characterCount={6} />
         </div>
       </div>
     </div>
--- a/apps/web/vibes/soul/primitives/products-carousel/index.tsx
+++ b/apps/web/vibes/soul/primitives/products-carousel/index.tsx
@@ -14,6 +14,7 @@ import {
   ProductCard,
   ProductCardSkeleton,
 } from '@/vibes/soul/primitives/product-card';
+import * as Skeleton from '@/vibes/soul/primitives/skeleton';
 
 export type CarouselProduct = CardProduct;
 
@@ -124,9 +125,13 @@ export function ProductsCarouselSkeleton({
   hideOverflow?: boolean;
 }) {
   return (
-    <div
-      className={clsx('@container', hideOverflow && 'overflow-hidden', className)}
-      data-pending={pending ? '' : undefined}
+    <Skeleton.Root
+      className={clsx(
+        'group-has-[[data-pending]]/products-carousel:animate-pulse',
+        hideOverflow && 'overflow-hidden',
+        className,
+      )}
+      pending={pending}
     >
       <div className="w-full">
         <div className="-ml-4 flex @2xl:-ml-5">
@@ -141,13 +146,13 @@ export function ProductsCarouselSkeleton({
         </div>
       </div>
       <div className="mt-10 flex w-full items-center justify-between gap-8">
-        <div className="h-1 w-full max-w-56 rounded bg-contrast-100" />
-        <div className="flex gap-2 text-contrast-200">
-          <ArrowLeft className="h-6 w-6" strokeWidth={1.5} />
-          <ArrowRight className="h-6 w-6" strokeWidth={1.5} />
+        <Skeleton.Box className="h-1 w-56 rounded" />
+        <div className="flex gap-2">
+          <Skeleton.Icon icon={<ArrowLeft aria-hidden className="h-6 w-6" strokeWidth={1.5} />} />
+          <Skeleton.Icon icon={<ArrowRight aria-hidden className="h-6 w-6" strokeWidth={1.5} />} />
         </div>
       </div>
-    </div>
+    </Skeleton.Root>
   );
 }
 
--- a/apps/web/vibes/soul/primitives/skeleton/index.tsx
+++ b/apps/web/vibes/soul/primitives/skeleton/index.tsx
@@ -1,0 +1,66 @@
+import { clsx } from 'clsx';
+import { ReactNode } from 'react';
+
+/**
+ * This component supports various CSS variables for theming. Here's a comprehensive list, along
+ * with their default values:
+ *
+ * ```css
+ * :root {
+ *   --skeleton: color-mix(in oklab, hsl(var(--contast-300)), white 75%);
+ * }
+ * ```
+ */
+function SkeletonRoot({
+  className,
+  children,
+  pending = false,
+}: {
+  className?: string;
+  children?: React.ReactNode;
+  pending: boolean;
+}) {
+  return (
+    <div
+      className={clsx('@container', className)}
+      data-pending={pending ? '' : undefined}
+      role={pending ? 'status' : undefined}
+    >
+      {children}
+      {pending && <span className="sr-only">Loading...</span>}
+    </div>
+  );
+}
+
+function SkeletonBox({ className }: { className?: string }) {
+  return <div className={clsx('bg-[var(--skeleton,hsl(var(--contrast-300)/15%))]', className)} />;
+}
+
+function SkeletonText({
+  characterCount = 10,
+  className,
+}: {
+  characterCount?: number | 'full';
+  className?: string;
+}) {
+  return (
+    <div className={clsx('flex h-[1lh] items-center', className)}>
+      <div
+        className={clsx(
+          `h-[1ex] max-w-full rounded-[inherit] bg-[var(--skeleton,hsl(var(--contrast-300)/15%))]`,
+          characterCount === 'full' ? 'w-full' : `w-[${characterCount}ch]`,
+        )}
+      />
+    </div>
+  );
+}
+
+function SkeletonIcon({ className, icon }: { className?: string; icon: ReactNode }) {
+  return (
+    <div className={clsx('text-[var(--skeleton,hsl(var(--contrast-300)))] opacity-25', className)}>
+      {icon}
+    </div>
+  );
+}
+
+export { SkeletonIcon as Icon, SkeletonRoot as Root, SkeletonBox as Box, SkeletonText as Text };
--- a/apps/web/vibes/soul/sections/featured-products-list/index.tsx
+++ b/apps/web/vibes/soul/sections/featured-products-list/index.tsx
@@ -29,6 +29,7 @@ export function FeaturedProductsList({
 }: Props) {
   return (
     <StickySidebarLayout
+      className="group/products-list"
       sidebar={
         <>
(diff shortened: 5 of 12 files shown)


apps/web/vibes/soul/primitives/skeleton/index.tsx
```diff
@@ -0,0 +1,135 @@
+import { clsx } from 'clsx';
+
+type Props = {
+  parentClassName?: string;
+  childClassName?: string;
+  rounded?: 'sm' | 'base' | 'md' | 'lg' | 'xl' | '2xl' | '3xl' | 'full';
+} & (
+  | { type: 'text'; characterCount: number; aspectRatio?: never }
+  | { type: 'image'; aspectRatio: '5:6' | '3:4' | '1:1'; characterCount?: never }
+  | { type?: 'box' | 'circle'; characterCount?: never; aspectRatio?: never }
```
apledger, line 10: `type: 'circle'` and `rounded: full` feel overloaded to me. Do you need a `type: 'circle'`?

apps/web/vibes/soul/primitives/skeleton/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+  rounded,
+  characterCount = 10,
+  aspectRatio = '5:6',
+}: Props) {
+  if (type === 'box') {
+    return (
+      <div className={clsx('flex items-center', parentClassName)}>
+        <div
+          className={clsx(
+            'bg-[var(--skeleton-background,hsl(var(--contrast-100)))]',
```
apledger, line 37: If there's no `--skeleton-text` or `skeleton-foreground` variable we can just call this `--skeleton`.

apps/web/vibes/soul/primitives/skeleton/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+            typeof rounded === 'string' &&
```
apledger, line 38: Why do you check for `string` type here?

apps/web/vibes/soul/primitives/skeleton/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+ * ```css
+ * :root {
+ *   --skeleton-background: hsl(var(--contrast-100));
+ * }
+ * ```
+ */
+
+export function Skeleton({
+  parentClassName,
+  childClassName,
```
apledger, line 26: What's the use case for exposing the `childClassName` here? It seems like the "parent" here is always just a `flex items-center` wrapper. The semantics of the "parent" vs the "child" here is unclear to me as a user of the component. I wonder if we can simplify this.

apps/web/vibes/soul/primitives/skeleton/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+            }[rounded],
+          childClassName,
+        )}
+        style={{ width: `${characterCount}ch` }}
+      />
+    </div>
+  );
+}
+
+export function SkeletonWrapper({
```
apledger, line 122: What you think of exporting this as `Skeleton.Root`, `Skeleton.Text`, `Skeleton.Image` etc instead of using a union grouped `type` prop?

apledger (commented):
I like this direction but think we need to refine the interface a bit. I'm wondering if we rethought the exports as `Skeleton.Root`, `Skeleton.Text`, `Skeleton.Image` etc instead how simple could the API look?

apps/web/vibes/soul/primitives/skeleton/index.tsx
hunterbecton: I like it!

apps/web/vibes/soul/primitives/skeleton/index.tsx
hunterbecton: The designs usually have a parent container that's the same height as the original component to minimize layout shift, but the actual skeleton inside typically has a smaller height than the parent container. I think the only time this is not the case is when the skeleton is for an image.

I agree that's not clear with `childClassName` and `parentClassName`. I wonder if there is a simpler approach I might not be thinking of? For example, I like that the `Skeleton.Text` is so simple because it's using `line-height` and `em` units to set the height automatically. I think we could do this with `Skeleton.Box` if we standardize how skeleton heights are set, for example, setting the height of the inner container to a percentage of the parent.

apps/web/vibes/soul/primitives/skeleton/index.tsx
hunterbecton: I was getting the following lint error without it: `Unsafe argument of type error typed assigned to a parameter of type `ClassValue`.eslint[@typescript-eslint/no-unsafe-argument](https://typescript-eslint.io/rules/no-unsafe-argument)`

apps/web/vibes/soul/primitives/product-card/index.tsx
```diff
@@ -164,19 +165,17 @@ export function ProductCard({
...
-            </div>
-            <div className="mb-2 flex h-[1lh] items-center text-sm font-normal text-contrast-400">
-              <span className="block h-[1ex] w-[8ch] rounded-sm bg-contrast-100" />
-            </div>
-            <div className="flex h-[1lh] items-center">
-              <span className="block h-[1ex] w-[5ch] rounded-sm bg-contrast-100" />
-            </div>
+            <Skeleton characterCount={10} rounded="base" type="text" />
+            <Skeleton characterCount={8} rounded="base" type="text" />
+            <Skeleton characterCount={5} rounded="base" type="text" />
```
migueloller, line 178: What if instead of `type` we had `Skeleton.Text`, `Skeleton.Image`, that way we can type the props based on the type. We don't want `characterCount` on `type` "image", etc.

apps/web/vibes/soul/primitives/product-card/index.tsx
```diff
@@ -164,19 +165,17 @@ export function ProductCard({
 export function ProductCardSkeleton({ className }: { className?: string }) {
   return (
     <div className={className}>
-      <div className="flex aspect-[5/6] flex-col gap-2 rounded-xl bg-contrast-100 @md:rounded-2xl" />
+      <Skeleton
+        aspectRatio="5:6"
+        childClassName="rounded-xl w-full h-full @md:rounded-2xl"
```
migueloller, line 170: This seems weird. Do we want an API that allows us to pass the class directly to the element?

apps/web/vibes/soul/primitives/skeleton/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+  children,
+  className,
+}: {
+  children: React.ReactNode;
+  className?: string;
+}) {
+  return (
+    <div className={clsx('motion-safe:animate-pulse', className)} role="status">
```
migueloller, line 130: I think that the issue with adding the pulse animation here is that we won't be able to change whether we pulse or not based on whether we're building an empty state or not. When I think of a "skeleton" it could be used both for loading or empty states, but whether it's pulsing depends on the pending state. Also, for situations like the PLP, the pending state is based on a sibling so we still want to use Tailwind `group` for that.

I like that we're encapsulating the accessibility, etc. but perhaps this component should be named `PendingWrapper` or something else? Also, do we want to have this sort of stuff in PLP? Do we need to go back to context so we can use `aria-hidden` and `sr-only` or is there a way we can do this with just CSS?

migueloller (dismissed):
I love this idea! Having a primitive to build skeletons seems great. I do think we some separation between skeleton and pending states, though, since we use skeletons for empty states as well and also we need to handle advanced pending patterns like we use in PLP

apps/web/vibes/soul/primitives/products-carousel/index.tsx
```diff
@@ -102,26 +103,28 @@ export function ProductsCarousel({
...
   className,
   placeholderCount = 8,
-  pending = false,
 }: {
   className?: string;
   placeholderCount?: number;
-  pending?: boolean;
 }) {
   return (
-    <Carousel className={className} data-pending={pending ? '' : undefined}>
```
migueloller, line 112: Don't we want to preserve the ability to pass in `pending` so that we can distinguish from a pending skeleton from a skeleton we use on an empty state (which isn't pending)?

apps/web/vibes/soul/primitives/skeleton/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+ * This component supports various CSS variables for theming. Here's a comprehensive list, along
+ * with their default values:
+ *
+ * ```css
+ * :root {
+ *   --skeleton: hsl(var(--contrast-100));
+ * }
+ * ```
+ */
+
```
migueloller, line 13: nit: remove space

apps/web/vibes/soul/primitives/skeleton/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+}: {
+  children: React.ReactNode;
+  className?: string;
+  pending?: boolean;
+}) {
+  return (
+    <div className={className} data-pending={pending ? '' : undefined} role="status">
+      <div
+        aria-hidden="true"
+        className="motion-safe:group-has-[[data-pending]]/pending:animate-pulse"
```
migueloller, line 135: I found myself doing groups that are more specific than `group/pending` when using the pattern. For example, I did `group/products-list-section` for the `<ProductsListSection>` component.

apps/web/vibes/soul/primitives/skeleton/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+        )}
+        style={{ width, height: characterHeight }}
+      />
+    </div>
+  );
+}
+
+SkeletonText.displayName = 'Skeleton.Text';
+
+function SkeletonPending({
```
migueloller, line 122: I'm still not following how this component is meant to be used in situations like `<ProductsListSection>` where the pending state isn't coupled to a skeleton. Or maybe it's not intended to be used there at all?

apps/web/vibes/soul/sections/featured-products-list/index.tsx
```diff
@@ -49,15 +50,13 @@ export function FeaturedProductsList({
       }
       sidebarSize="1/3"
     >
-      <div className="group-has-[[data-pending]]/pending:animate-pulse">
```
migueloller, line 52: Why remove this? Now, when switching filters in PLP, there's no pending state

migueloller (commented):
The new skeleton components look good! What I still don't quite follow is how we're handling pending states. We seem to be removing functionality we had for the products list section where we showed a pending state on transitions (completely independent of using skeletons)

apps/web/vibes/soul/primitives/card/index.tsx
```diff
@@ -133,11 +133,7 @@ export function CardSkeleton({
           }[aspectRatio],
         )}
       />
-      <div className="mt-3 text-lg">
-        <div className="flex h-[1lh] items-center">
-          <span className="block h-[1ex] w-[10ch] rounded-sm bg-contrast-100" />
-        </div>
-      </div>
+      <div className="mt-3 text-lg">{/* TODO: Add card skeleton */}</div>
```
jamesqquick, line 136: Do we need to finish this todo?

apps/web/vibes/soul/examples/sections/products-carousel/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <div>
+      <SectionLayout className="group/products-carousel">
+        <ProductsCarousel
+          emptyStateSubtitle="Try browsing our complete catalog of products."
+          emptyStateTitle="No products found"
+          products={products}
+        />
+      </SectionLayout>
+      <SectionLayout className="group/products-carousel bg-foreground">
```
jamesqquick, line 17: Is `group/products-carousel` a Tailwind syntax? What do we call that?

apps/web/vibes/soul/primitives/skeleton/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+ * ```
+ */
+function SkeletonRoot({
+  className,
+  children,
+  pending = false,
+}: {
+  className?: string;
+  children?: React.ReactNode;
+  pending: boolean;
```
jamesqquick, line 20: Are skeletons not always pending?

apps/web/vibes/soul/primitives/card/index.tsx
hunterbecton: I was thinking of just doing it as part of my card review. I didn't want to use the new skeleton component here until it was approved.

apps/web/vibes/soul/examples/sections/products-carousel/index.tsx
hunterbecton: Yeah, it's basically a way to give a name to a group so we can target specific pending styles to them. If we just called all of the groups `group/pending` and have nested classes with the same name, it could break the pattern we're using to handle when skeletons should animate.

apps/web/vibes/soul/primitives/skeleton/index.tsx
hunterbecton: The idea is that skeletons could also be used in empty states, so we'd not want them to animate. The `pending` prop also handles some accessibility stuff like rendering `{pending && <span className="sr-only">Loading...</span>}` and `role=status`.

jamesqquick: Ohhhh ok got it. Thanks!

apps/web/vibes/soul/examples.ts
migueloller, line 1: I was a bit thrown off that these changes from primitives to sections were in this PR

apps/web/vibes/soul/sections/section-layout/index.tsx
```diff
@@ -17,15 +17,13 @@ export function SectionLayout({
...
-  hideOverflow = false,
 }: {
   className?: string;
   children: React.ReactNode;
   containerSize?: 'md' | 'lg' | 'xl' | '2xl';
-  hideOverflow?: boolean;
 }) {
   return (
-    <section className={clsx('@container', hideOverflow && 'overflow-hidden', className)}>
+    <section className={clsx(`@container`, className)}>
```
migueloller, line 26: Why use back ticks if we're not interpolating?

apps/web/vibes/soul/primitives/product-card/index.tsx
```diff
@@ -183,18 +184,10 @@ export function ProductCardSkeleton({
...
-              <span className="block h-[1ex] w-[8ch] rounded-sm bg-contrast-100" />
-            </div>
-            <div className="flex h-[1lh] items-center">
-              <span className="block h-[1ex] w-[5ch] rounded-sm bg-contrast-100" />
-            </div>
-          </div>
+        <div className="w-full text-base">
+          <Skeleton.Text className="w-[10ch] rounded" />
+          <Skeleton.Text className="w-[8ch] rounded" />
+          <Skeleton.Text className="w-[5ch] rounded" />
```
migueloller, line 190: I think that for the `Skeleton.Text` component there might be some value in passing in `characterCount` as a prop to hide the "internals" of using the `ch` unit. What do you think?

migueloller: Also, we probably have a standard amount of "roundedness" given the size the text, no?

migueloller (commented):
Changes look good but this PR seems to be mixing carousel changes with skeleton changes. For future PRs, I'd appreciate if we could break this down into 2 PRs 🙏🏻 

Also, I don't love adding placeholders like we are here in the same PR where we're introducing the feature. Why not go ahead and use the component? It feels "incomplete" to merge. Perhaps that should be in a separate draft PR?

migueloller (approved):

apps/web/vibes/soul/primitives/product-card/index.tsx
hunterbecton: I went back and forth about exposing a `characterCount` prop. I think I decided to go against it because the inline-styles would override the Tailwind class if you wanted the skeleton text to be another size that's not character width. I'm still open to exploring that though.

There is a roundedness that's associated with the size of the text. Should I create a `size` prop to handle these styles automatically?

migueloller: Yeah, I was thinking that we'd want a standard size and character count for this particular component. We can use it as-is for a while and see how it feels

hunterbecton: I added `characterCount` back, with the option to set it to full-width. Setting a high character count value didn't fill the space like I expected, so having a full width option seemed necessary.

Ultimately, I want to create typography component that handles the resizing and rounded automatically, but that will be a future release.

jamesqquick (approved):

PR #439, opened 2024-12-23
feat: create BlogPostContentSkeleton, make BlogPostContent props streamable

jordanarldt:
## Summary
- Create BlogPostContentSkeleton component do display while blog post page is being loaded
- Make the BlogPostContent props streamable in order to display the skeleton before the data is ready

## Mobile

image

image

## Desktop

image

image

### DIFF
--- a/apps/web/vibes/soul/primitives/blog-post-card/index.tsx
+++ b/apps/web/vibes/soul/primitives/blog-post-card/index.tsx
@@ -2,7 +2,7 @@ import { clsx } from 'clsx';
 import Image from 'next/image';
 import Link from 'next/link';
 
-export interface BlogPost {
+export interface BlogPostCardBlogPost {
   id: string;
   author?: string | null;
   content: string;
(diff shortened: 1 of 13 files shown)


apps/web/vibes/soul/primitives/blog-post-card/index.tsx
```diff
@@ -16,7 +16,15 @@ export interface BlogPost {
...
-export function BlogPostCard({ title, image, content, href, date, author, className }: BlogPost) {
+export function BlogPostCard({
+  title,
+  image,
+  content,
+  href,
+  date,
+  author,
+  className,
+}: BlogPostCardBlogPost) {
```
migueloller, line 27: I think we should avoid having an interface drive props. Instead we should do this:

```ts
function BlogPostCard({ blogPost }: { blogPost: BlogPostCardBlogPost }) {
```

Alternatively, we could just have these be props and there's no reason to share the interface, we could share the prop interface. If you think about it from a UI perspective, the `title` doesn't necessarily have to come from a blog post—for example, in Makeswift, you could create a "fake" `BlogPostCard` that has this data provided visually.

A good example of where this breaks down is that the `id` field is required for a "blog post" but we don't want to require `id` for this component as a prop. We went through this for the `ProductCard` component

migueloller (changes requested):
I think these are mostly looking good but the one thing I think we should change is to decouple the UI from the data. i.e., the blog card component's props should not be coupled to the idea of a blog post

migueloller (approved):
Looks great 👍🏻

PR #440, opened 2024-12-29
feat: pending states with Tailwind group

migueloller:
Instead of using React context, which relies on JS, we can handle pending states using Tailwind `group` and `group-has-{mofifier}` classes.


apps/web/vibes/soul/sections/products-list-section/index.tsx
```diff
@@ -62,103 +60,101 @@ export function ProductsListSection({
...
-                        <FiltersPanel
-                          filters={filters}
-                          paginationInfo={paginationInfo}
-                          resetFiltersLabel={resetFiltersLabel}
-                        />
-                      </Suspense>
-                    </SidePanel.Content>
-                  </SidePanel.Root>
-                </div>
+    <div className="group/products-list-section @container">
```
migueloller, line 63: Enable "Hide whitespace" to view a better diff that shows the important changes

CleanShot 2024-12-29 at 14 42 39@2x

apps/web/vibes/soul/form/select/index.tsx
```diff
@@ -11,6 +11,7 @@ import { Label } from '@/vibes/soul/form/label';
 type Props = {
   id?: string;
   name: string;
+  pending?: boolean;
```
migueloller, line 14: Had to add a `pending` prop to `Select` because we spread to Radix's `Select.Root` which doesn't render a DOM element.

PR #444, opened 2024-12-30
feat: add Error page and section

jordanarldt:
Add the `Error` page component and examples

## Screenshots
image

### DIFF
--- a/apps/web/vibes/soul/docs/page-error.mdx
+++ b/apps/web/vibes/soul/docs/page-error.mdx
@@ -1,0 +1,5 @@
+---
+title: Error Page
+preview: page-error-example
+previewSize: xl
+---
--- a/apps/web/vibes/soul/examples/pages/error/index.tsx
+++ b/apps/web/vibes/soul/examples/pages/error/index.tsx
@@ -1,0 +1,87 @@
+import { locales } from '@/vibes/soul/data/locales';
+import { action } from '@/vibes/soul/examples/primitives/inline-email-form/actions';
+import { localeAction } from '@/vibes/soul/examples/primitives/navigation/actions';
+import { logo, navigationLinks } from '@/vibes/soul/examples/primitives/navigation/electric';
+import { copyright, footerLinks } from '@/vibes/soul/examples/sections/footer/electric';
+import { Banner } from '@/vibes/soul/primitives/banner';
+import { Navigation } from '@/vibes/soul/primitives/navigation';
+import { Error } from '@/vibes/soul/sections/error';
+import { Footer } from '@/vibes/soul/sections/footer';
+import {
+  Amex,
+  ApplePay,
+  Bitcoin,
+  GooglePay,
+  Mastercard,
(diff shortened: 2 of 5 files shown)


apps/web/vibes/soul/docs/page-error.mdx
andrewreifman, line 1: I think we only need 1 example page for this view. We only need to have multiple examples when we need to change the content within each view, like the products displayed in the Not Found examples.

The brand/style switcher already handles changing colors and typography within a single example where the content doesn't change (see Blog page).

apps/web/vibes/soul/sections/error/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <section className="@container">
+      <div className="mx-auto max-w-3xl px-4 py-10 @xl:px-6 @xl:py-14 @4xl:px-8 @4xl:py-20">
+        <h1 className="mb-3 font-heading text-3xl font-medium leading-none @xl:text-4xl @4xl:text-5xl">
+          {title}
+        </h1>
+        <p className="text-lg text-contrast-500">{subtitle}</p>
+
+        {cta !== undefined && cta.href !== '' && cta.label !== '' && (
+          <ButtonLink className="mt-8" href={cta.href} size="large" type="button" variant="primary">
```
andrewreifman, line 28: Should we use the `Button` component here instead of `ButtonLink`? I assume the action a user will take is something that will either reload the view or something similar?

apps/web/vibes/soul/docs/page-error.mdx
jordanarldt: Good call-out, thanks for this!

apps/web/vibes/soul/sections/error/index.tsx
andrewreifman: We're deciding to keep the ButtonLink for now

apps/web/vibes/soul/examples/pages/error/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+        activeLocaleId="en"
+        cartHref="#"
+        links={navigationLinks}
+        localeAction={localeAction}
+        locales={locales}
+        logo={logo}
+        searchHref="#"
+      />
+
+      <Error cta={{ label: 'Try again', href: '#' }} />
```
andrewreifman, line 70: Since we already have default prop data included in the Error component, does it make sense to add the CTA prop data there as well so it's all in one place?

jordanarldt: @andrewreifman I was thinking the same thing at first, but I'm not privy to the idea of defaulting to `#` href that does nothing 🤔  The way it is currently, if `cta` prop is not provided at all, the button won't render. Otherwise, it will always use the default button, and the user would have to explicitly provide `undefined` if they don't want the button to show.

andrewreifman (approved):

PR #445, opened 2024-12-31
feat: update Error to use click event handler

jordanarldt:
I noticed NextJS requires Error page to be a client component, so we need to use an onClick handler instead of an href.

Updated `Error` section to allow a function to be passed in rather than a string.

https://nextjs.org/docs/app/building-your-application/routing/error-handling#using-error-boundaries

### DIFF
--- a/apps/web/vibes/soul/examples/pages/error/index.tsx
+++ b/apps/web/vibes/soul/examples/pages/error/index.tsx
@@ -48,6 +48,14 @@ const paymentIconsArray: React.ReactNode[] = [
   <Bitcoin key="Bitcoin" />,
 ];
 
+async function ctaAction() {
+  'use server';
+
+  await new Promise<void>((resolve) => {
+    setTimeout(() => resolve(), 1000);
+  });
+}
+
 export default function Preview() {
   return (
     <>
@@ -67,7 +75,7 @@ export default function Preview() {
         searchHref="#"
       />
 
-      <Error cta={{ label: 'Try again', href: '#' }} />
+      <Error ctaAction={ctaAction} />
 
       <Subscribe
         action={action}
--- a/apps/web/vibes/soul/sections/error/index.tsx
+++ b/apps/web/vibes/soul/sections/error/index.tsx
@@ -1,20 +1,17 @@
-import { ButtonLink } from '@/vibes/soul/primitives/button-link';
-
-interface Link {
-  label: string;
-  href: string;
-}
+import { Button } from '@/vibes/soul/primitives/button';
 
 interface Props {
   title?: string;
   subtitle?: string;
-  cta?: Link;
+  ctaLabel?: string;
+  ctaAction?: () => void | Promise<void>;
 }
 
 export function Error({
   title = 'Something went wrong!',
   subtitle = 'Please try again or contact our support team for assistance.',
-  cta,
+  ctaLabel = 'Try again',
+  ctaAction,
 }: Props) {
   return (
     <section className="@container">
@@ -24,10 +21,12 @@ export function Error({
         </h1>
         <p className="text-lg text-contrast-500">{subtitle}</p>
 
-        {cta !== undefined && cta.href !== '' && cta.label !== '' && (
-          <ButtonLink className="mt-8" href={cta.href} size="large" type="button" variant="primary">
-            {cta.label}
-          </ButtonLink>
+        {ctaAction && (
+          <form action={ctaAction}>
+            <Button className="mt-8" size="large" type="submit" variant="primary">
+              {ctaLabel}
+            </Button>
+          </form>
         )}
       </div>
     </section>
--- a/apps/web/.eslintrc.json
+++ b/apps/web/.eslintrc.json
@@ -34,6 +34,7 @@
     "@typescript-eslint/no-shadow": "error",
     "@typescript-eslint/no-unused-vars": ["error", { "ignoreRestSiblings": true }],
     "@typescript-eslint/no-unnecessary-condition": "error",
+    "@typescript-eslint/no-misused-promises": ["error", { "checksVoidReturn": { "attributes": false } }],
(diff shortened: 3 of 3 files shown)


apps/web/vibes/soul/sections/error/index.tsx
```diff
@@ -1,14 +1,14 @@
...
   label: string;
-  href: string;
+  onRetry: () => void;
 }
 
 interface Props {
   title?: string;
   subtitle?: string;
-  cta?: Link;
+  cta?: Action;
```
migueloller, line 11: Given the use of the word "Action" for server actions and client actions I think we should avoid this name

apps/web/vibes/soul/sections/error/index.tsx
```diff
@@ -1,14 +1,14 @@
-import { ButtonLink } from '@/vibes/soul/primitives/button-link';
+import { Button } from '@/vibes/soul/primitives/button';
 
-interface Link {
+interface Action {
   label: string;
-  href: string;
+  onRetry: () => void;
```
migueloller, line 5: nit: I'd rename this `onClick` since `onRetry` is too specific. There's no retry logic in this component

apps/web/vibes/soul/sections/error/index.tsx
migueloller: How about we take in a `ctaLabel` and `ctaOnClick` props?

Also, what if we always rendered the CTA button using a default label and opted for `window.location.refresh()` if a value isn't provided?

apps/web/vibes/soul/sections/error/index.tsx
```diff
@@ -1,20 +1,23 @@
...
-  href: string;
-}
+import { Button } from '@/vibes/soul/primitives/button';
 
 interface Props {
   title?: string;
   subtitle?: string;
-  cta?: Link;
+  ctaLabel?: string;
+  ctaOnClick?: () => void;
```
migueloller, line 7: Let's rename this to `ctaAction`

apps/web/vibes/soul/sections/error/index.tsx
```diff
@@ -1,20 +1,23 @@
...
+}
+
+function defaultOnClick() {
+  if (typeof window !== 'undefined') {
```
migueloller, line 11: There should be no need to check for `typeof window !== 'undefined'` since click handlers can only run in a environment with a DOM

apps/web/vibes/soul/sections/error/index.tsx
```diff
@@ -24,10 +27,12 @@ export function Error({
         </h1>
         <p className="text-lg text-contrast-500">{subtitle}</p>
 
-        {cta !== undefined && cta.href !== '' && cta.label !== '' && (
-          <ButtonLink className="mt-8" href={cta.href} size="large" type="button" variant="primary">
-            {cta.label}
-          </ButtonLink>
+        {ctaLabel !== '' && (
```
migueloller, line 30: I don't think we should check for the empty string here. If somebody explicitly passes that, it's kind of on them if we render a button with an empty label

migueloller (changes requested):

apps/web/vibes/soul/examples/pages/error/index.tsx
```diff
@@ -67,7 +67,7 @@ export default function Preview() {
         searchHref="#"
       />
 
-      <Error cta={{ label: 'Try again', href: '#' }} />
+      <Error ctaAction={() => null} />
```
migueloller, line 70: Let's not pass anything. This is causing the build to fail. Another option would be to pass in 

```ts
async () => {
  'use server'

  // do nothing
}
```

apps/web/vibes/soul/sections/error/index.tsx
```diff
@@ -1,20 +1,21 @@
...
   title?: string;
   subtitle?: string;
-  cta?: Link;
+  ctaLabel?: string;
+  ctaAction?: () => void;
+}
+
+function defaultAction() {
+  window.location.reload();
 }
```
migueloller, line 12: I actually think this default action might be a problem since this file doesn't have `'use client'`. Any RSC that imports it would run into an issue if it doesn't add `use client`. An easy way around it would be to have a separate file for the form but we could also simply remove the default action and do nothing on click

migueloller (changes requested):

apps/web/vibes/soul/sections/error/index.tsx
jordanarldt: @migueloller I agree, instead of doing nothing on click i think it would make sense to simply not display the button if no action is provided

apps/web/vibes/soul/sections/error/index.tsx
```diff
@@ -24,10 +21,12 @@ export function Error({
...
-        {cta !== undefined && cta.href !== '' && cta.label !== '' && (
-          <ButtonLink className="mt-8" href={cta.href} size="large" type="button" variant="primary">
-            {cta.label}
-          </ButtonLink>
+        {ctaAction && (
+          <form action={ctaAction}>
+            <Button className="mt-8" size="large" type="submit" variant="primary">
+              {ctaLabel}
+            </Button>
+          </form>
```
jorgemoya, line 29: Do we need a form? Per the [docs](
@bigcommerce/catalyst-core:dev: [BigCommerce] query GetFavicon - 84ms - complexity 1002
) a simple `onClick` handler should work.

jordanarldt: @jorgemoya We don't need a form since NextJS error pages are always client components, but @migueloller suggested using a form for versatility - just in-case someone might want to use the component in a non-standard setting

jorgemoya (approved):

migueloller (approved):

PR #447, opened 2025-01-02
fix: slideshow typo max-w-x -> max-w-xl

fikrikarim:


apledger (approved):
Thanks for catching this @fikrikarim.

PR #448, opened 2025-01-02
Cart upgrades

apledger:
- [x] Refactored Cart to use `Streamable` pattern
- [x] Polished skeletons to perfectly match all breakpoints
- [x] Refactored `Summary` props to allow for arbitrary line items


apps/web/vibes/soul/sections/cart/index.tsx
```diff
@@ -25,18 +27,9 @@ export interface CartLineItem {
...
-  subtotalLabel?: string;
-  subtotal: string | Promise<string>;
-  shippingLabel?: string;
-  shipping?: string;
-  taxLabel?: string;
-  tax?: string | Promise<string>;
-  grandTotalLabel?: string;
-  grandTotal?: string | Promise<string>;
-  ctaLabel?: string;
+interface CartSummaryLineItem {
```
migueloller, line 30: A line item is wording usually used in a receipt/order/checkout so I think it makes sense for `cart.lineItems` since those have a quantity and a price, etc. For the summary, I think these are genera UI list items so `CartSummaryItem` is probably fine

apps/web/vibes/soul/sections/cart/index.tsx
```diff
@@ -53,71 +46,62 @@ interface CartState<LineItem extends CartLineItem> {
...
   title?: string;
-  lineItems: LineItem[] | Promise<LineItem[]>;
-  summary: CartSummary;
-  emptyState: CartEmptyState;
+  summaryTitle?: string;
+  emptyState?: CartEmptyState;
   lineItemAction: Action<CartState<LineItem>, FormData>;
   checkoutAction: Action<SubmissionResult | null, FormData>;
+  checkoutLabel?: string;
+  summaryTotalLabel?: string;
```
migueloller, line 62: What do you think if we had a `summary` prop and it had to fields, `items` and `subtotal` and `items` was a `CartSummaryItem[]` and `subtotal` was `CartSummaryItem` but we render it with a different visual style?

apps/web/vibes/soul/sections/cart/index.tsx
```diff
@@ -53,71 +46,62 @@ interface CartState<LineItem extends CartLineItem> {
...
 interface CartProps<LineItem extends CartLineItem> {
   title?: string;
-  lineItems: LineItem[] | Promise<LineItem[]>;
-  summary: CartSummary;
-  emptyState: CartEmptyState;
+  summaryTitle?: string;
+  emptyState?: CartEmptyState;
   lineItemAction: Action<CartState<LineItem>, FormData>;
   checkoutAction: Action<SubmissionResult | null, FormData>;
+  checkoutLabel?: string;
```
migueloller, line 61: I think we're going to need to split this component into two files because we want the labels to be streamable as well since those are localized. Then, we'll want the component that uses `Stream` to be a RSC so we can `Promise.all` all streamable values

apps/web/vibes/soul/sections/cart/index.tsx
```diff
@@ -378,90 +350,94 @@ function SubmitButton(props: React.ComponentPropsWithoutRef<typeof Button>) {
...
-          <div className="mt-10 h-[50px] w-full rounded-full bg-contrast-100" />
-        </>
+          <div className="mt-4 h-[58px] w-full rounded-full bg-contrast-100" />
+        </div>
       }
       sidebarPosition="after"
       sidebarSize="1/3"
     >
-      <div className="w-full">
+      <div data-pending={pending ? '' : undefined}>
```
migueloller, line 404: We only need the `data-pending` at the top level

migueloller (approved):
Looks pretty good to me 👍🏻 

Just had a couple of comments

apps/web/vibes/soul/sections/cart/index.tsx
apledger: I did this because the `StickySidebarLayout` component doesn't spread props. Do you think we should spread on that component?

apps/web/vibes/soul/sections/cart/index.tsx
apledger: Sounds good. Why would we stream in a labels instead of just awaiting them btw?

apps/web/vibes/soul/sections/cart/index.tsx
apledger: `subtotal` seems to be rendered as just a regular summary item though, why separate it? Are you maybe getting it confused with `total` which does need to be separated? I went back and forth on this for a while and figured that grouping things into `summary` was unnecessary. The concept of a `total` price is represented in different ways on different carts and may not necessarily be in the summary section. Nesting things inside of `summary` makes an assumption about the UI which I felt would potentially cause awkwardness down the road in other VIBES.

migueloller: I thought "total" was only shown in the checkout view and that the cart only had a subtotal, but yeah, I meant the item that was itself styled a bit differently than the others. I wasn't thinking of the summary section when I read the summary prop and was thinking more about the summary of the cart concept itself, not any particular UI. It just feels a bit inconsistent what we've been nesting and what we haven't been nesting. It's not a big deal, though

apps/web/vibes/soul/sections/cart/index.tsx
migueloller: It's what we've been doing so far for localized labels. Perhaps it's unnecessary and this will all probably change once we implement label i18n

apps/web/vibes/soul/sections/cart/index.tsx
migueloller: We only need it in 1 place and we're using it in the sidebar we're passing to `StickySidebarLayout`

hunterbecton (approved):
LGTM!

PR #451, opened 2025-01-06
fix: force default value for inputs in PDP

jorgemoya:
Even though we generate default values:

```ts
const defaultValue = fields.reduce<{
  [Key in keyof SchemaRawShape]?: z.infer<SchemaRawShape[Key]>;
}>(
  (acc, field) => ({
    ...acc,
    [field.name]: params[field.name] ?? field.defaultValue ?? '',
  }),
  { quantity: 1 },
);
```

Example of defaultValue object:

```
 {
  '108': '99',
  '109': '106',
  '130': '',
  '133': '137',
  '134': '140',
  '135': 'false',
  '136': '',
  '137': '',
  '138': '',
  '140': '',
  quantity: 1
}
```

For some reason, `useForm` sets empty strings as `undefined` for `defaultValue/initialValue`. Problem is we control all these fields in this form, so we can't change from `uncontrolled => controlled` without a warning.

Basically, it transforms to:

```
 {
  '108': '99',
  '109': '106',
  '130': undefined,
  '133': '137',
  '134': '140',
  '135': 'false',
  '136': undefined,
  '137': undefined,
  '138': undefined,
  '140': undefined,
  quantity: 1
}
```

With this change, we make sure the form field remains in a controlled state.

Why is `useForm` swapping `''` to `undefined`, this I couldn't figure out. It does make sense that it works this way since I assume many uses cases are for uncontrolled forms. Actual non empty string values do remain and work as expected.

### DIFF
--- a/apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
+++ b/apps/web/vibes/soul/sections/product-detail/product-detail-form.tsx
@@ -220,7 +220,7 @@ function FormField({
           onChange={(e) => handleChange(e.currentTarget.value)}
           onFocus={controls.focus}
           required={formField.required}
-          value={controls.value}
+          value={controls.value ?? ''}
         />
       );
 
@@ -235,7 +235,7 @@ function FormField({
           onChange={(e) => handleChange(e.currentTarget.value)}
           onFocus={controls.focus}
           required={formField.required}
-          value={controls.value}
+          value={controls.value ?? ''}
         />
       );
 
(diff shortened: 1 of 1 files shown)


jorgemoya:
> This is weird behavior for sure. Did you try upgrading to see if that fixed it? Or perhaps check open issues?

Couldn't find anything with a quick glance but I will investigate more before merging.

jorgemoya:
@migueloller Found this issue and response from mantainer: https://github.com/edmundhung/conform/issues/802

Seems like something they will fix eventually.

PR #454, opened 2025-01-08
feat: add CSS theme variables for form components

hunterbecton:
## What/Why
- Adds new CSS variables to support custom theming for forms

## Testing

https://github.com/user-attachments/assets/cd85bdaf-d2ab-49a8-9266-d989c6ed6caf

### DIFF
--- a/apps/web/vibes/soul/form/form-status/index.tsx
+++ b/apps/web/vibes/soul/form/form-status/index.tsx
@@ -1,21 +1,52 @@
 import { clsx } from 'clsx';
 import { CheckCircle, CircleAlert } from 'lucide-react';
 
+/**
+ * This component supports various CSS variables for theming. Here's a comprehensive list, along
+ * with their default values:
+ *
+ * ```css
+ *  :root {
+ *    --form-status-light-background-error: color-mix(in oklab, hsl(var(--error)), white 75%);
+ *    --form-status-light-text-error: color-mix(in oklab, hsl(var(--error)), black 75%);
+ *    --form-status-light-background-success: color-mix(in oklab, hsl(var(--success)), white 75%);
+ *    --form-status-light-text-success: color-mix(in oklab, hsl(var(--success)), black 75%);
+ *    --form-status-dark-background-error: color-mix(in oklab, hsl(var(--error)), white 75%);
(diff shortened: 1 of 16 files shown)


migueloller (dismissed):
Changes look good! I'll let @apledger review the actual variables, the naming, etc.

apps/web/vibes/soul/form/form-status/index.tsx
```diff
@@ -1,21 +1,52 @@
...
+ * ```
+ */
 export function FormStatus({
   className,
   children,
   type = 'success',
+  colorScheme = 'light',
   ...rest
-}: React.ComponentPropsWithoutRef<'div'> & { type?: 'error' | 'success' }) {
+}: React.ComponentPropsWithoutRef<'div'> & {
```
jamesqquick, line 27: Do we need to update this to not use the introspection for props?

apps/web/vibes/soul/form/input/index.tsx
```diff
@@ -4,23 +4,53 @@ import * as React from 'react';
...
+ *   --input-dark-border: hsl(var(--contrast-500));
+ *   --input-dark-focus: hsl(var(--background));
+ *   --input-dark-placeholder: hsl(var(--contrast-100));
+ *   --input-dark-error: hsl(var(--error));
+ *  }
+ * ```
+ */
 export const Input = React.forwardRef<
   React.ComponentRef<'input'>,
   Omit<React.ComponentPropsWithoutRef<'input'>, 'id'> & {
```
jamesqquick, line 30: Same question about inferred props

apps/web/vibes/soul/form/form-status/index.tsx
jamesqquick: I guess this would be the other option?

```
React.HTMLAttributes<HTMLDivElement>
```

If you're using the pattern that was already there, go for it.

jamesqquick (approved):

apps/web/vibes/soul/form/form-status/index.tsx
hunterbecton: Yeah, I actually think it makes a lot of sense to use that here, but maybe not when we're passing in a component. This was already here, but not that you've pointed it out, I need to make a pass at these form components again for the changes we're doing now. These updates are kinda stale.

PR #456, opened 2025-01-08
feat(soul): refactor order detail page

chanceaclark:
### Updates `shipments` to `destinations`

This change is due to how ecommerce orders are actually handled. For example, when you place an order you can ship X items to one address, and ship Y items to another. This is also called consignments, but to make it more friendly to the end user, I opted for the name `destinations` instead.

### Multiple shipments per destination.

In ecomm, each destination can have multiple shipments. This all depends on how the order is fulfilled by each merchant. Some products need to be sent in multiple boxes (e.g. furniture) or send in partial shipments (e.g. as products arrive at the warehouse). I refactored the schema to handle multiple shipments under each destination.

### Tracking information link vs. text

Added the ability to handle links in the tracking information field. With this came renaming `id` to `number` as it might cause some confusion for developers that they might need an entity id instead.

### Internationalization

Elevated some hard coded strings into props to provide better internationalization to strings.

## Testing
Screenshot 2025-01-08 at 14 45 44

### DIFF
--- a/apps/web/vibes/soul/examples/sections/order-details-section/electric.tsx
+++ b/apps/web/vibes/soul/examples/sections/order-details-section/electric.tsx
@@ -6,15 +6,13 @@ const products = await getProducts('Electric');
 
 const order = {
   id: '1',
-  title: 'Order #1',
   status: 'Delivered',
   statusColor: 'success' as const,
   date: '2021-01-01',
-  shipments: [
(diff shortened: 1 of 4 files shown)


jorgemoya (approved):
lgtm!

migueloller (approved):
Looks good! 👍🏻

PR #457, opened 2025-01-08
feat: add sr-only heading for additional information in PDP

jorgemoya:
Fixes this issue in PDP:

image

### DIFF
--- a/apps/web/vibes/soul/sections/product-detail/index.tsx
+++ b/apps/web/vibes/soul/sections/product-detail/index.tsx
@@ -39,6 +39,7 @@ interface Props<F extends Field> {
   ctaDisabled?: Streamable<boolean | null>;
   prefetch?: boolean;
   thumbnailLabel?: string;
+  additionalInformationLabel?: string;
 }
 
 export function ProductDetail<F extends Field>({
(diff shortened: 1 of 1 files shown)


apps/web/vibes/soul/sections/product-detail/index.tsx
```diff
@@ -137,6 +139,7 @@ export function ProductDetail<F extends Field>({
                     }
                   </Stream>
 
+                  <h2 className="sr-only">{additionalInformationLabel}</h2>
```
migueloller, line 142: 🍹 Should we name this `additionalInformationTitle` instead?

Also, is "additional information" the right way to frame this? Or is there a different term like "product specifications"?

migueloller (dismissed):

apps/web/vibes/soul/sections/product-detail/index.tsx
jorgemoya: I struggled with this, technically it can be used for anything, but yeah product specifications might fit the bill better.

jorgemoya: Actually, in our case, we render product specifications and warranty information, which is why I went with "additional information". I'm okay with changing to `additionaInformationTitle`.

migueloller (approved):

PR #460, opened 2025-01-09
Add Down for Maintenance page

jordanarldt:
Add Down for Maintenance page to vibes
image

### DIFF
--- a/apps/web/vibes/soul/sections/maintenance/index.tsx
+++ b/apps/web/vibes/soul/sections/maintenance/index.tsx
@@ -1,0 +1,81 @@
+import { MailIcon, PhoneIcon } from 'lucide-react';
+import Link from 'next/link';
+
+import { Logo } from '@/vibes/soul/primitives/logo';
+import { SectionLayout } from '@/vibes/soul/sections/section-layout';
+
+interface Image {
+  src: string;
+  alt: string;
+}
+
+interface Props {
+  className?: string;
+  logo?: string | Image | null;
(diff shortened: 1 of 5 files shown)


apps/web/vibes/soul/sections/maintenance/index.tsx
```diff
@@ -0,0 +1,59 @@
+import { clsx } from 'clsx';
+import Link from 'next/link';
+
+interface Props {
+  className?: string;
+  logo?: React.ReactNode;
```
apledger, line 6: I think this should be the same `logo` prop type that's used in the `Navigation` and `Footer` that uses the `Logo` component.

jordanarldt: @apledger Fair enough!

apps/web/vibes/soul/sections/maintenance/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+  className = '',
+  logo,
+  title = 'We are down for maintenance',
+  statusMessage = "Sorry for the inconvenience, we're currently working on improving our store.",
+  contactText = 'Contact us',
+  contactPhone,
+  contactEmail,
+}: Props) {
+  return (
+    <section className={clsx('@container', className)}>
```
apledger, line 32: Can you use `SectionLayout` here instead?

chanceaclark (approved):

PR #461, opened 2025-01-10
Add first iteration of Web Pages

jordanarldt:
Add the first iteration of Web Pages component to vibes.

The next iteration will contain a new layout and support for nested web page children / links.

## Testing

image


jorgemoya:
I feel like this is just a Catalyst thing and we don't need to create a section for this.

jordanarldt:
> I feel like this is just a Catalyst thing and we don't need to create a section for this.

Makes sense to me! I'm down for either way.

PR #464, opened 2025-01-10
Lazy load embla carousels

joshuawootonn:


migueloller (dismissed):
LGTM 👍🏻

Something that would be nice is if we could render an LQIP instead of nothing at all when the image isn't in view. Perhaps something to tackle in the next iteration.

PR #468, opened 2025-01-10
Add Contact Form Section

jordanarldt:
Add `ContactFormSection` to VIBES


apps/web/vibes/soul/sections/dynamic-form-section/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+        <header className="mx-auto w-full max-w-4xl pb-8 @2xl:pb-12 @4xl:pb-16">
+          <h1 className="mb-5 font-heading text-4xl font-medium leading-none @xl:text-5xl">
+            {title}
+          </h1>
+          {Boolean(subtitle) && subtitle !== '' && (
+            <p className="mb-10 text-base font-light leading-none @xl:text-lg">{subtitle}</p>
+          )}
+        </header>
+      )}
+      <div className="mx-auto w-full max-w-4xl">
```
apledger, line 34: Do we need this div? Max width is being set by the section layout.

jordanarldt: @apledger Seems like it! Without it, the alignment is off:
image

migueloller: It's because we're using `mx-auto` to center the contents. Since `SectionLayout` takes `className` we could use Tailwind there to center the contents by using flexbox or any other centering method. We could also leave this div here but then we can probably drop the `max-w-4xl` class

apps/web/vibes/soul/sections/dynamic-form-section/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+  action,
+}: Props<F>) {
+  return (
+    <SectionLayout className={className} containerSize="lg">
+      {Boolean(title) && (
+        <header className="mx-auto w-full max-w-4xl pb-8 @2xl:pb-12 @4xl:pb-16">
+          <h1 className="mb-5 font-heading text-4xl font-medium leading-none @xl:text-5xl">
+            {title}
+          </h1>
+          {Boolean(subtitle) && subtitle !== '' && (
```
migueloller, line 29: I think we should, as a general rule, avoid coercing values using constructors of primitives, like `Boolean`. Instead, we should be explicit and check `subtitle != null` which will cover `null` and `undefined`.

migueloller: This also goes for the usage for `title` above

jordanarldt: @migueloller Hmm, that is an interesting approach. I've always went by the opposite, that strict equal `===` `!==` should always be preferred over the simple equal. Is there a reason why we would prefer that over type coercion with the `Boolean` constructor? The Boolean constructor seems like a better way to capture what we intend with less code. In fact, using `Boolean(subtitle)` technically means the `subtitle !== ''` is also unnecessary:

> `Boolean(null) = false`
> `Boolean(undefined) = false`
> `Boolean('') = false`

apps/web/vibes/soul/sections/dynamic-form-section/index.tsx
jordanarldt: @migueloller Hmm. Whenever I make those changes, it works well in VIBES, but the child `div` in the SectionLayout component has padding by default (and can't be overridden by props), so it still messes up the layout:

image

jordanarldt: I guess in Catalyst we can just not use the `DynamicFormSection` component and stick to the `DynamicForm` by itself 👍

jorgemoya (approved):

apps/web/vibes/soul/sections/dynamic-form-section/index.tsx
migueloller: If the intent is to handle falsy values like `''`, then yes, I agree with you that it is more succinct than having to do `subtitle != null & subtitle !== ''`. That's good w/ me. As for always preferring `===` over `==`, that's a rule that I always follow, except when checking for both `null` and `undefined`. Instead of `val !== null && val !== undefined`, I opt for `val != null` since that's equivalent. I think the same logic applies to `Boolean(val)`. In that case, though, you could also do `!val` or `!!val`. I guess `Boolean(val)` feels more "explicit" and we can reserve `!` only for boolean values.

PR #473, opened 2025-01-28
fix: issue with skeleton height in Header

jorgemoya:
- Fixes an issue with the height of the link skeletons and the resulting links.
- Updated the skeletons to be a little longer depicting more real use cases.
- Fix a linting issues that we get in Catalyst.

## Before
https://github.com/user-attachments/assets/f250f89b-923b-421a-afc7-203a7a6ff17f

## After
https://github.com/user-attachments/assets/7cdad347-9d1f-4997-992c-9823861a1742

### DIFF
--- a/apps/web/vibes/soul/primitives/navigation/index.tsx
+++ b/apps/web/vibes/soul/primitives/navigation/index.tsx
@@ -307,7 +307,7 @@ export const Navigation = forwardRef(function Navigation<S extends SearchResult>
     >
       <div
         className={clsx(
-          'flex items-center justify-between gap-1 bg-[var(--nav-background,hsl(var(--background)))] py-2 pl-3 pr-2 transition-shadow @4xl:rounded-2xl @4xl:px-2 @4xl:pl-6 @4xl:pr-2.5',
+          'flex min-h-16 items-center justify-between gap-1 bg-[var(--nav-background,hsl(var(--background)))] py-2 pl-3 pr-2 transition-shadow @4xl:rounded-2xl @4xl:px-2 @4xl:pl-6 @4xl:pr-2.5',
           isFloating
             ? 'shadow-xl ring-1 ring-[var(--nav-floating-border,hsl(var(--foreground)/10%))]'
             : 'shadow-none ring-0',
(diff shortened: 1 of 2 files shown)


apps/web/vibes/soul/primitives/navigation/index.tsx
```diff
@@ -420,18 +420,18 @@ export const Navigation = forwardRef(function Navigation<S extends SearchResult>
...
-                  <span className="block h-4 w-14 rounded-md bg-contrast-100" />
+                  <span className="block h-5 w-12 rounded-md bg-contrast-100" />
                 </li>
                 <li>
-                  <span className="block h-4 w-24 rounded-md bg-contrast-100" />
+                  <span className="block h-5 w-24 rounded-md bg-contrast-100" />
                 </li>
                 <li>
-                  <span className="block h-4 w-16 rounded-md bg-contrast-100" />
+                  <span className="block h-5 w-16 rounded-md bg-contrast-100" />
```
migueloller, line 434: @hunterbecton, we probably want to replace this with our latest skeleton patterns where we using the appropriate CSS units for text, etc.

I wouldn't block this PR for this, though, since we haven't finalized those

migueloller (approved):

jorgemoya:
The height is off by 1px. Need to fix.

apps/web/vibes/soul/primitives/navigation/index.tsx
```diff
@@ -420,12 +420,12 @@ export const Navigation = forwardRef(function Navigation<S extends SearchResult>
         >
           <Stream
             fallback={
-              <ul className="flex animate-pulse flex-row p-2 @4xl:gap-2 @4xl:p-5">
+              <ul className="flex animate-pulse flex-row gap-6 p-[12.5px]">
```
jorgemoya, line 423: Reason why I set `12.5px` is because the height of the header with links rendered is 73px minimum. With this we can match the exact height, but I'm open to other suggestions.

Screenshot 2025-01-28 at 2 14 43 PM

migueloller: I think generally, the approach with this is that the header should have a defined height, then the content inside is centered vertically. This way, instead of the height being driven by the contents, it's predictable. Now, IIRC we're letting the logo drive the height after a certain point. So we want to make sure that we're using a logo and a skeleton logo that's below that threshold so that the minimum height of the header is the driving value here

jorgemoya: Good callout. I updated to set a `min-height` to the header, and remove paddings from the skeletons. There will still be a jump when the logo renders, but that will be expected/we should let the user modify their min-height if their logo is too large.

migueloller: Is there a jump in our demo store? I feel like within a reasonable size of logo there should be no jump at all

migueloller (dismissed):

apps/web/vibes/soul/primitives/navigation/index.tsx
jorgemoya: Not for us with these changes.

migueloller (approved):

PR #474, opened 2025-01-29
fix: remove carousel from skeletons

hunterbecton:
## What / Why

- Remove Embla carousel from skeleton and empty components
- Add streamable pattern to `CardCarousel`

### Skeletons

https://github.com/user-attachments/assets/68737111-f905-4fbd-97e7-5d615e58a9b5

### Empty state

https://github.com/user-attachments/assets/6f8d2945-31e5-46c7-a221-ef9e6481c030

### DIFF
--- a/apps/web/vibes/soul/primitives/card-carousel/index.tsx
+++ b/apps/web/vibes/soul/primitives/card-carousel/index.tsx
@@ -1,5 +1,7 @@
 import { clsx } from 'clsx';
+import { ArrowLeft, ArrowRight } from 'lucide-react';
 
+import { Stream, Streamable } from '@/vibes/soul/lib/streamable';
 import { Card, CardProps, CardSkeleton } from '@/vibes/soul/primitives/card';
 import {
   Carousel,
@@ -8,8 +10,6 @@ import {
   CarouselItem,
   CarouselScrollbar,
 } from '@/vibes/soul/primitives/carousel';
-
-import { Stream, Streamable } from '../../lib/streamable';
 
 export type Card = CardProps & {
   id: string;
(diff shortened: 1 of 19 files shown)


apps/web/vibes/soul/primitives/card-carousel/index.tsx
```diff
@@ -22,7 +22,8 @@ export interface CardCarouselProps {
   iconColorScheme?: 'light' | 'dark';
   carouselColorScheme?: 'light' | 'dark';
   className?: string;
-  emptyStateMessage?: string;
+  emptyStateTitle?: Streamable<string | null>;
+  emptyStateSubtitle?: Streamable<string | null>;
```
migueloller, line 26: I think for these ones we just want `Streamable<string>` since if we didn't have streamable we wouldn't have `string | null`. This is a departure from an original rule we had where any optional prop would include `| null`. We realized this was a bad rule and that we should only make it nullable if that would be the case without `Streamable`.

migueloller: Something else... I'm not sure we should use `Streamable` in our primitives and reserve it just for our sections. It feels like primitives should be, well, primitive and not include logic for pending states. Instead, the skeleton components are almost primitives themselves. The section components can use the skeletons together with the primitives and `<Stream>` to make it all work together. What do you think?

migueloller (approved):
LGTM. Got a couple of comments, but nothing to block the PR

apps/web/vibes/soul/primitives/card-carousel/index.tsx
hunterbecton: That makes a lot of sense to me. I thought it felt weird to have streamable in primitives, but wasn't sure if I was missing something. 😅 

I'm going to add these notes for when we implement streamable patterns.

PR #475, opened 2025-01-29
DON'T MERGE - Testing Inline Data for Cart Component

jamesqquick:
Testing to see what it looks like to inline data in examples instead of using shared data and actions. This allows examples to be copy/pastable.


migueloller (dismissed):
This looks good to me. I think we can address the repetition as a fast-follow after we release Soul

PR #478, opened 2025-01-31
Inline data for product primitive examples

jamesqquick:
- inline data in `<Products-Carousel>` and `<Products-List>` components
-  shorten example names to remove `-example`
- add separate examples for `<Products-Carousel>` to show differentiated data
- removed loader examples and added loading skeletons/streaming into primary example


apps/web/vibes/soul/examples/primitives/products-carousel/electric.tsx
```diff
@@ -0,0 +1,185 @@
...
+  subtitle: string;
+  badge?: string;
+  price: string;
+  image: {
+    src: string;
+    alt: string;
+  };
+  href: string;
+  rating: number;
+}
```
migueloller, line 36: Don't we export this type from `products-carousel`?

migueloller (dismissed):
Looks good 👍🏻 

One minor thing I noticed is that we're re-defining the `Product` type on each example and I was thinking we might be able to reuse the ones from the sections

apps/web/vibes/soul/examples/primitives/products-carousel/electric.tsx
jamesqquick: Great catch. Removed these duplicate types and used the ones that come from the primitive component

migueloller (approved):

PR #479, opened 2025-01-31
Inline Data for Cart Section

jamesqquick:
- inline line item data for Cart
- simplify example names
- remove loading examples (will circle back with these when we fix TypeScript issue)
- add cookie logic to have line items and summary details update appropriately as well as persist on reloads in example
- add "Reset Cart" button to reset back to default data


apps/web/vibes/soul/examples/sections/cart/electric.tsx
```diff
@@ -1,24 +1,247 @@
...
+interface Cart<CartLineItem> {
+  id: string;
+  lineItems: CartLineItem[];
+  summaryItems: SummaryItems;
+  total: string;
+}
+
+/**
+ * Replace these with your own implementation
+ */
```
migueloller, line 148: This JSDoc comment is going to be associated with `getCart` but it seems to be intended for the rest of the file, not just this function. I'd suggest not using JSDoc and instead using a regular multi-line comment

apps/web/vibes/soul/examples/sections/cart/electric.tsx
```diff
@@ -1,24 +1,247 @@
...
-      lineItemAction={lineItemAction}
-    />
+    <div className="p-4">
+      <Form action={resetCart}>
+        <Button type="submit" variant={'ghost'}>
+          Reset cart
+        </Button>
+      </Form>
+      <Cart
+        cart={await getCart()}
```
migueloller, line 20: Do we want to stream this in for the example?

migueloller (approved):
Looks good. Got some minor comments that apply to all three examples

migueloller:
I think I prefer the class as a way to group all related logic together for the data access layer

apps/web/vibes/soul/examples/sections/cart/electric.tsx
jamesqquick: This was the one I was having the TypeScript issue with. I couldn't figure out how to get around the TypeScript issues :(

jamesqquick:
@migueloller I can go back and add that!

migueloller:
Don't feel like you have to. But I do feel like these examples are starting to get pretty verbose and I think that organizing things that way might make it easier to figure out what's going on

apps/web/vibes/soul/examples/sections/cart/electric.tsx
migueloller: I left a comment in Slack on how to address it by using the `Omit` utility on the `Cart` props. Let me know if you're still having trouble with it after applying that change

apps/web/vibes/soul/examples/sections/cart/electric.tsx
jamesqquick: Went ahead and removed it since, I think it's obvious they'll need their own implementation

jamesqquick:
@migueloller decided to hold off on moving back to class just to keep the changes simple for now.

hunterbecton (approved):

PR #480, opened 2025-01-31
Inline Data for `<Featured-Products-Carousel>` Examples

jamesqquick:
- inline data in `Featured-Products-Carousel`
- add example for each style with distinct data
- simplify example names

https://github.com/user-attachments/assets/e761b091-f2cc-432b-a382-129a8656b3ca


hunterbecton:
This LGTM. Since we're inline the data in the examples now, is there a process you're following so things are easier to keep up with? For example, would we still have a single data file from which we'd copy and paste from?

hunterbecton (commented):
This LGTM! Since we're inline the data in the examples now, is there a process you're following so things are easier to keep up with? For example, would we still have a single data file from which we'd copy and paste from?

hunterbecton (dismissed):

hunterbecton (approved):

migueloller (commented):
I'm late to the party but this looks good! 👍🏻

PR #481, opened 2025-01-31
`<Featured-Products-List>` Inline Data

jamesqquick:
- inline data in `<Featured-Products-List>`
- add example for each style with distinct data
- simplify example names
- update data references in Home page component

https://github.com/user-attachments/assets/95546a54-f50d-4c45-b357-3ad39bcbde9e


apps/web/vibes/soul/examples/pages/home/electric.tsx
```diff
@@ -270,3 +269,164 @@ export default function Preview() {
...
+    href: '#',
+    rating: 4.4,
+  },
+  {
+    id: '3',
+    title: 'Pink Caladium',
+    subtitle: 'Indoor Plant',
+    price: '$19.95',
+    image: {
+      src: 'https://rstr.in/monogram/vibes/AaZW4j2VTd4',
```
hunterbecton, line 306: We already discussed this in our stand-up, but I think we need to update these src image links to be the same host and also include blurDataURL. I created an issue here: https://linear.app/makeswift/issue/VIB-3970/standardize-where-we-import-image-provide-blurdataurl

hunterbecton (approved):
LGTM! I added a small comment about the host for our images, but we already discussed that in standup and I created an issue to address it.

PR #484, opened 2025-02-04
chore: enable format on save

migueloller:
We had reports that files weren't being formatted on save so I'm adding
this option for a consistent experience for VS Code users.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,6 @@
 {
   "typescript.tsdk": "node_modules/typescript/lib",
   "eslint.workingDirectories": [{ "mode": "auto" }],
-  "editor.formatOnSave": true
+  "editor.formatOnSave": true,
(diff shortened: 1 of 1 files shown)


jamesqquick:
It doesn't appear to work on each file. For me, it worked in component `.tsx` files, but it didn't work in `.ts` files.

https://github.com/user-attachments/assets/45fd65fe-bf84-4bca-b751-20f0d1f890da

jamesqquick:
When I try to run the format document command from command palette it gives me this `There are multiple formatters for 'TypeScript' files. One of them should be configured as default formatter.`
 
CleanShot 2025-02-04 at 14 31 02@2x

jamesqquick:
@migueloller It worked!!! YAYYYYYYY :)

jamesqquick (approved):

PR #488, opened 2025-02-06
Hunter/vib 3952 initial code audit group 1

hunterbecton:
# What/Why

This PR updates the following components:
- `Accordion`
- `Alert`
- `AnimatedLink`
- `Badge`
- `Banner`
- `BlogPostCard`
- `BlogPostCarousel`
- `BlogList`
- `Breadcrumbs`
- `Button`
- `ButtonLink`
- `Calendar`
- `Card`
- `CardCarousel`
- `Carousel`
- `Checkbox`

Why so many file changes for such a short list of components? Well, because the changes affected examples and other components across the project. I outline below why these changes are important, and we should decide which ones we want to carry over.

### `label` to `text`
I opted to rename `label` to `text` throughout VIBES as I believe `label` should be reserved for label elements. For example, our `Checkbox` component takes a `label` prop, which is passed into a Radix `Label` component. 

### More descriptive props
I think being more descriptive with some props, like changing `image` on the `BlogPostCard` to `thumbnail` is a better experience for the user because the image's intent is clearer.

### More selective to when we use `children`
It makes sense for components like `Accordion` because users should be able to render whatever they want inside the accordion content. 

However, I think we should consider not using `children` inside components like `Button,` `ButtonLink,` and similar. 
Instead, we should design the component to support different styles, like rendering an icon through props, rather than just passing the icon or other components to `children.` 

I know libraries like Radix don't do this with their components, but Soul is much different. Soul's design is opinionated, and I think our components should be, too.

## Some notes

None of the components above have their skeleton or empty state updated. Those updates will be completed when this PR is merged (making some changes now based on Miguel's latest feedback): https://github.com/makeswift/vibes/pull/436

Updates to the `Calendar` component do not include any of the UX fixes from this issue: https://linear.app/makeswift/issue/VIB-1059/datepicker-ux-improvements

### Requiring an `id` in our data

You'll see in the PR that I'm requiring an `id` for data that will be rendered from arrays, for example:

```
Array<ComponentPropsWithoutRef<typeof AnimatedLink> & { id: string }>
```

I know this is how I would do it from my projects because I know each element would have a `uuid` in my database, however, should we require an `id`?

### DIFF
--- a/apps/web/vibes/soul/primitives/accordions/index.tsx
+++ b/apps/web/vibes/soul/primitives/accordions/index.tsx
@@ -2,7 +2,7 @@
 
 import * as AccordionsPrimitive from '@radix-ui/react-accordion';
 import { clsx } from 'clsx';
-import { useEffect, useState } from 'react';
+import { ComponentPropsWithoutRef, useEffect, useState } from 'react';
 
 /**
  * This component supports various CSS variables for theming. Here's a comprehensive list, along
@@ -10,11 +10,14 @@ import { useEffect, useState } from 'react';
  *
  * ```css
  * :root {
+ *   --accordion-focus: hsl(var(--primary));
+ *   --acordion-light-offset: hsl(var(--background));
  *   --accordion-light-title-text: hsl(var(--contrast-400));
  *   --accordion-light-title-text-hover: hsl(var(--foreground));
  *   --accordion-light-title-icon: hsl(var(--contrast-500));
  *   --accordion-light-title-icon-hover: hsl(var(--foreground));
  *   --accordion-light-content-text: hsl(var(--foreground));
+ *   --acordion-dark-offset: hsl(var(--foreground));
  *   --accordion-dark-title-text: hsl(var(--contrast-200));
  *   --accordion-dark-title-text-hover: hsl(var(--background));
  *   --accordion-dark-title-icon: hsl(var(--contrast-200));
@@ -26,11 +29,12 @@ import { useEffect, useState } from 'react';
  * ```
  */
 function Accordion({
-  children,
   title,
+  children,
   colorScheme = 'light',
-  ...rest
-}: React.ComponentPropsWithoutRef<typeof AccordionsPrimitive.Item> & {
+  className,
+  ...props
+}: ComponentPropsWithoutRef<typeof AccordionsPrimitive.Item> & {
   colorScheme?: 'light' | 'dark';
 }) {
   const [isMounted, setIsMounted] = useState(false);
@@ -40,12 +44,22 @@ function Accordion({
   }, []);
 
   return (
-    <AccordionsPrimitive.Item {...rest}>
+    <AccordionsPrimitive.Item
+      {...props}
+      className={clsx(
+        'focus:outline-2 has-[:focus-visible]:ring-2 has-[:focus-visible]:ring-[var(--accordion-focus,hsl(var(--primary)))] has-[:focus-visible]:ring-offset-4',
+        {
+          light: 'ring-offset-[var(--acordion-light-offset,hsl(var(--background)))]',
+          dark: 'ring-offset-[var(--acordion-dark-offset,hsl(var(--foreground)))]',
+        }[colorScheme],
+        className,
+      )}
+    >
       <AccordionsPrimitive.Header>
-        <AccordionsPrimitive.Trigger className="group flex w-full cursor-pointer items-start gap-8 py-3 text-start @md:py-4">
+        <AccordionsPrimitive.Trigger className="group flex w-full cursor-pointer items-start gap-8 border-none py-3 text-start focus:outline-none @md:py-4">
           <div
             className={clsx(
-              'flex-1 select-none font-[family-name:var(--accordion-title-font-family,var(--font-family-mono))] text-sm uppercase transition-colors duration-300 ease-out',
+              'flex-1 select-none font-[family-name:var(--accordion-title-font-family,var(--font-family-mono))] text-sm font-normal uppercase transition-colors duration-300 ease-out',
               {
                 light:
                   'text-[var(--accordion-light-title-text,hsl(var(--contrast-400)))] group-hover:text-[var(--accordion-light-title-text-hover,hsl(var(--foreground)))]',
@@ -76,7 +90,7 @@ function Accordion({
       >
         <div
           className={clsx(
-            'pb-5 font-[family-name:var(--accordion-content-font-family)] font-medium leading-normal',
+            'py-3 font-[family-name:var(--accordion-content-font-family,var(--font-family-body))] text-base font-light leading-normal',
             {
               light: 'text-[var(--accordion-light-content-text,hsl(var(--foreground)))]',
               dark: 'text-[var(--accordion-dark-content-text,hsl(var(--background)))]',
@@ -92,11 +106,11 @@ function Accordion({
 
 function AnimatedChevron({
   className,
-  ...rest
+  ...props
 }: React.JSX.IntrinsicAttributes & React.SVGProps<SVGSVGElement>) {
   return (
     <svg
-      {...rest}
+      {...props}
       className={clsx(
         'mt-1 shrink-0 [&>line]:origin-center [&>line]:transition [&>line]:duration-300 [&>line]:ease-out',
         className,
--- a/apps/web/vibes/soul/primitives/alert/index.tsx
+++ b/apps/web/vibes/soul/primitives/alert/index.tsx
@@ -1,21 +1,25 @@
 import { clsx } from 'clsx';
 import { X } from 'lucide-react';
-import { ReactNode } from 'react';
 
 import { Button } from '@/vibes/soul/primitives/button';
 
-interface Props {
-  variant: 'success' | 'warning' | 'error' | 'info';
-  message: ReactNode;
-  description?: string;
-  dismissLabel?: string;
-  action?: {
-    label: string;
-    onClick: () => void;
-  };
-  onDismiss?: () => void;
-}
-
+/**
+ * This component supports various CSS variables for theming. Here's a comprehensive list, along
+ * with their default values:
+ *
+ * ```css
+ * :root {
+ *   --alert-success-background: color-mix(in oklab, hsl(var(--success)), white 75%);
+ *   --alert-warning-background: color-mix(in oklab, hsl(var(--warning)), white 75%);
+ *   --alert-error-background: color-mix(in oklab, hsl(var(--error)), white 75%);
+ *   --alert-info-background: hsl(var(--background));
+ *   --alert-font-family: var(--font-family-body);
+ *   --alert-border: hsl(var(--foreground) / 10%);
+ *   --alert-message-text: hsl(var(--foreground));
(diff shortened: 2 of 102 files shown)


migueloller:
> ## Changes
> ### `label` to `text`
> I opted to rename `label` to `text` throughout VIBES as I believe `label` should be reserved for label elements. For example, our `Checkbox` component takes a `label` prop, which is passed into a Radix `Label` component.

I don't like using the word "text" for props because it lacks semantic meaning. It's similar to naming a prop `image` instead of `thumbnail` or `coverImage`, etc. I prefer "label" but I also understand how it can be confusing to use that word since there's the `<label>` HTML element. You have a section below on "More descriptive props" and I think this change runs a bit counter to that. Again, maybe the conflict with `<label>` warrants it, but just wanted to share my 2c.

> ### More selective to when we use `children`
> It makes sense for components like `Accordion` because users should be able to render whatever they want inside the accordion content.
> 
> However, I think we should consider not using `children` inside components like `Button,` `ButtonLink,` and similar. Instead, we should design the component to support different styles, like rendering an icon through props, rather than just passing the icon or other components to `children.`
> 
> I know libraries like Radix don't do this with their components, but Soul is much different. Soul's design is opinionated, and I think our components should be, too.

I think we should be careful here not to conflate the `children` prop with a prop of type `ReactNode`. We can use any type for the `children` prop. That being said, I think that we need to be practical here. I feel like locking things down too much in primitives, like Button, might make development cumbersome. I ran into this firsthand in our web app where we decided to have `iconLeft` and `iconRight` props instead of just allowing `children` as a `ReactNode`. It's something I regret and we're not going back to the Radix UI & shadcn/ui way of doing things.

> ### Requiring an `id` in our data
> You'll see in the PR that I'm requiring an `id` for data that will be rendered from arrays, for example:
> 
> ```
> Array<ComponentPropsWithoutRef<typeof AnimatedLink> & { id: string }>
> ```
> 
> I know this is how I would do it from my projects because I know each element would have a `uuid` in my database, however, should we require an `id`?

This one is a bit tricky. Requiring an ID for an array prop that won't necessarily come from a database lends itself for the user to pass indices as IDs. I think this will be more of a case by case basis thing but we should be careful to apply this as a blanket rule, IMO.

apps/web/vibes/soul/primitives/accordions/index.tsx
```diff
@@ -40,12 +44,22 @@ function Accordion({
...
-    <AccordionsPrimitive.Item {...rest}>
+    <AccordionsPrimitive.Item
+      {...props}
+      className={clsx(
+        'focus:outline-2 has-[:focus-visible]:ring-2 has-[:focus-visible]:ring-[var(--accordion-focus,hsl(var(--primary)))] has-[:focus-visible]:ring-offset-4',
+        {
+          light: 'ring-offset-[var(--acordion-light-offset,hsl(var(--background)))]',
+          dark: 'ring-offset-[var(--acordion-dark-offset,hsl(var(--foreground)))]',
+        }[colorScheme],
+        className,
```
migueloller, line 55: Have we considered that if we use `clsx` like this to concatenate classes we probably also need `tailwind-merge`? I know @apledger's not a huge fan of it but if we're going down the route of allow to override classes, we should probably use it

apps/web/vibes/soul/primitives/alert/index.tsx
```diff
@@ -1,65 +1,81 @@
...
   onDismiss,
-}: Props) {
+}: {
+  variant: 'success' | 'warning' | 'error' | 'info';
+  message: string;
+  description?: string;
+  dismissLabel?: string;
+  action?: {
+    onClick: () => void;
+    text: string;
```
migueloller, line 37: I don't love using `text` here. Specially if the rationale is we're trying to avoid a conflict with `<label>` because it couples the prop interface to the internal implementation details. For example, imagine this `action` prop was actually used with a `<form>` as a React Action and so `action.text` was actually now used as a `<label>`? The idea that we're making decisions about the prop interface based on internal implementation details doesn't sit well with me

apps/web/vibes/soul/primitives/animated-link/index.tsx
```diff
@@ -1,25 +1,36 @@
...
+export function AnimatedLink({
+  className,
+  text,
+  href,
+}: {
   className?: string;
-}
-
-export function AnimatedLink({ link, label, className }: AnimatedLinkProps) {
+  text: string;
```
migueloller, line 22: I think this should be `children: string`

apps/web/vibes/soul/primitives/banner/index.tsx
```diff
@@ -17,71 +17,51 @@ import { ForwardedRef, forwardRef, ReactNode, useCallback, useEffect, useState }
...
-      const hidden = localStorage.getItem(`${id}-hidden-banner`) === 'true';
-
-      setBanner({ dismissed: hidden, initialized: true });
-    }, [id]);
-
-    const hideBanner = useCallback(() => {
-      setBanner((prev) => ({ ...prev, dismissed: true }));
-      localStorage.setItem(`${id}-hidden-banner`, 'true');
-      onDismiss?.();
-    }, [id, onDismiss]);
```
migueloller, line 52: Are we just dropping this logic? Catalyst currently depends on this so we need to make sure we have an alternative approach to persisting hiding the banner. If could be handled outside of this component, but we should probably put that in an example

apps/web/vibes/soul/primitives/blog-post-card/index.tsx
```diff
@@ -2,66 +2,86 @@ import { clsx } from 'clsx';
...
+ *   --blog-post-card-summary-text: hsl(var(--contrast-400));
+ *   --blog-post-card-author-date-text: hsl(var(--foreground));
+ * }
+ * ```
+ */
+export function BlogPostCard({
+  blogPost,
+  className,
+}: {
+  blogPost: {
```
migueloller, line 26: I still feel like this component shouldn't have a `blogPost` prop as it implies we're making a decision on what a "blog post" entity should look like. The properties of `blogPost` should be at the top level props, IMO

apps/web/vibes/soul/primitives/blog-post-card/index.tsx
```diff
@@ -2,66 +2,86 @@ import { clsx } from 'clsx';
...
-            month: 'long',
-            day: 'numeric',
-          })}
-        </time>
-        {date !== '' && author != null && author !== '' && (
-          <span className="after:mx-2 after:content-['•']" />
-        )}
-        {author != null && author !== '' && <span>{author}</span>}
-      </div>
+      <article>
```
migueloller, line 49: Is `<article>` the appropriate HTML element for a card?

apps/web/vibes/soul/primitives/blog-post-carousel/index.tsx
```diff
@@ -7,21 +9,23 @@ import {
...
   previousLabel,
   nextLabel,
-}: Props) {
+}: {
+  className?: string;
+  blogPosts: Array<
+    ComponentPropsWithoutRef<typeof BlogPostCard>['blogPost'] & {
+      id: string;
+    }
+  >;
```
migueloller, line 24: I don't like this. I feel like we should export the props instead and alias them when imported. Or decouple the props and define them here

apps/web/vibes/soul/primitives/blog-post-carousel/index.tsx
```diff
@@ -7,21 +9,23 @@ import {
...
   blogPosts,
   scrollbarLabel,
   previousLabel,
   nextLabel,
-}: Props) {
+}: {
+  className?: string;
+  blogPosts: Array<
+    ComponentPropsWithoutRef<typeof BlogPostCard>['blogPost'] & {
+      id: string;
```
migueloller, line 22: What if we named this `key`?

apps/web/vibes/soul/primitives/breadcrumbs/index.tsx
```diff
@@ -1,20 +1,31 @@
...
+ *   --breadcrumbs-icon: hsl(var(--contrast-500));
+ *   --breadcrumbs-hover: hsl(var(--primary));
+ * }
+ * ```
+ */
+export function Breadcrumbs({
+  breadcrumbs: streamableBreadcrumbs,
+  className,
+}: {
+  breadcrumbs: Streamable<Array<ComponentPropsWithoutRef<typeof AnimatedLink> & { id: string }>>;
```
migueloller, line 26: I _really_ dislike doing this here as well. The API surface area of the breadcrumbs should be even more limited than `AnimatedLink`. I think we need to define what a "breadcrumb item" is, completely independently of `AnimatedLink`. `AnimatedLink` is an implementation detail and here we're coupling it with the breadcrumb items

migueloller, line 26: This is one of those situations where I'm not sure there will always be an ID and perhaps the best thing is to have no `key`. Also, I think I would name it `key` instead of `id`.

apps/web/vibes/soul/primitives/card-carousel/index.tsx
migueloller, line 1: I think this component should be a section

apps/web/vibes/soul/primitives/blog-post-carousel/index.tsx
migueloller, line 1: I think this component should be a section

apps/web/vibes/soul/primitives/checkbox/index.tsx
migueloller, line 1: I think that this component should be either renamed to `CheckboxField` or be broken up into `Checkbox` and `CheckboxField`

apps/web/vibes/soul/primitives/compare-card/index.tsx
migueloller, line 1: I think it's a bit weird that we have a separate component for comparisons. IMO, this primitive shouldn't exist and `ProductCard` should have optional capacity of compare functionality

apps/web/vibes/soul/primitives/products-list/compare-drawer.tsx
migueloller, line 1: I feel like this component should be a section

apps/web/vibes/soul/primitives/toaster/index.tsx
```diff
@@ -32,7 +31,7 @@ export const Toaster = ({ ...props }: ToasterProps) => {
 };
 
 export const toast = {
-  success: (message: ReactNode, options?: ToastOptions) => {
```
migueloller, line 35: There's a reason why we had support for `ReactNode` here. I think we need to make sure we look at usage in Catalyst before breaking the APIs. I'm pretty sure Catalyst relies on being able to pass a `ReactNode` here. I think we can change the API so that not everything takes a `ReactNode`, but as it is right now, this would break Catalyst

migueloller (commented):
Ok, just finished taking a look. I limited myself to only reviewing `apps/web/vibes/soul/primitives/*` files.

This is definitely too big of a PR. I think that a PR per primitive would be more digestible and would allow us to incrementally get there vs having an all-or-nothing PR like this one. It would also be better to keep discussions scoped to a particular component. Is that the plan or was the plan to merge this PR?

PR #491, opened 2025-02-10
audit: Alert component

hunterbecton:
## What/Why
- Adds CSS variables
- Exports interface

## Testing

https://github.com/user-attachments/assets/d0c4dc32-5088-42b9-99ec-d8a8a216c184


apps/web/vibes/soul/primitives/alert/index.tsx
```diff
@@ -16,41 +16,63 @@ interface Props {
...
+            'bg-[var(--alert-error-background,color-mix(in_oklab,_hsl(var(--error)),_white_75%))]',
+          info: 'bg-[var(--alert-info-background,hsl(var(--background)))]',
         }[variant],
       )}
       role="alert"
     >
-      <div className="flex flex-col">
-        <span className="text-sm font-normal text-foreground">{message}</span>
+      <div className="font-[family-name:var(--alert-font-family,var(--font-family-body))]">
+        <h5 className="text-sm font-normal text-[var(--alert-message-text,hsl(var(--foreground)))]">
```
jamesqquick, line 61: Should we be making an assumption as to what heading level this is? I'm not sure `h5` is going to provide semantic html since we don't know what other levels it will be surrounded by on the page? Did you see this pattern somewhere else?

hunterbecton: I noticed ShadCN does it on their `Alert` [component](https://ui.shadcn.com/docs/components/alert), but I agree we're assuming it makes sense in every context. I'm fine with using a `span`.

jamesqquick: Interesting. Ok, feels reasonable to copy theirs

jamesqquick (approved):

PR #505, opened 2025-02-12
feat: add coupon code form to Cart

jorgemoya:
Adds coupon code form to Cart section.

In coupon code you can:
1. Apply a coupon code
2. Remove existing coupon codes

Exposed props to Cart section to pass to the form. Updated examples to include coupon code form.

## Demo
https://github.com/user-attachments/assets/4d5e020a-7696-4977-94ce-da2d166a45cc

## Error state
Screenshot 2025-02-12 at 2 24 26 PM

### DIFF
--- a/apps/web/vibes/soul/examples/sections/cart/actions.ts
+++ b/apps/web/vibes/soul/examples/sections/cart/actions.ts
@@ -1,7 +1,11 @@
 import { SubmissionResult } from '@conform-to/react';
 import { parseWithZod } from '@conform-to/zod';
 
-import { cartLineItemActionFormDataSchema } from '@/vibes/soul/sections/cart/schema';
+import { CouponCodeFormState } from '@/vibes/soul/sections/cart/coupon-code-form';
+import {
+  cartLineItemActionFormDataSchema,
+  couponCodeActionFormDataSchema,
+} from '@/vibes/soul/sections/cart/schema';
 
 interface CartLineItem {
   id: string;
@@ -86,4 +90,42 @@ export async function checkoutAction(
   // return redirect('/checkout')
 
   return prevState;
+}
+
+export async function couponCodeAction(
+  prevState: Awaited<CouponCodeFormState>,
+  formData: FormData,
+): Promise<CouponCodeFormState> {
+  'use server';
+
+  const submission = parseWithZod(formData, { schema: couponCodeActionFormDataSchema });
+
+  if (submission.status !== 'success') {
+    return {
+      ...prevState,
+      lastResult: submission.reply({ formErrors: ['This code has expired or is invalid!'] }),
+    };
+  }
+
+  // Simulate a network request
+  await new Promise((resolve) => setTimeout(resolve, 1000));
+
+  switch (submission.value.intent) {
+    case 'apply': {
+      const couponCode = submission.value.couponCode;
+
+      return {
+        couponCodes: [...prevState.couponCodes, couponCode],
+        lastResult: submission.reply({ resetForm: true }),
+      };
+    }
+    case 'delete': {
+      const couponCode = submission.value.couponCode;
+
+      return {
+        couponCodes: prevState.couponCodes.filter((code) => code !== couponCode),
+        lastResult: submission.reply({ resetForm: true }),
+      };
+    }
+  }
 }
--- a/apps/web/vibes/soul/examples/sections/cart/actions.ts
+++ b/apps/web/vibes/soul/examples/sections/cart/actions.ts
@@ -103,7 +103,7 @@ export async function couponCodeAction(
   if (submission.status !== 'success') {
     return {
       ...prevState,
-      lastResult: submission.reply({ formErrors: ['This code has expired or is invalid!'] }),
+      lastResult: submission.reply(),
     };
   }
 
--- a/apps/web/vibes/soul/examples/sections/cart/actions.ts
+++ b/apps/web/vibes/soul/examples/sections/cart/actions.ts
@@ -98,7 +98,7 @@ export async function couponCodeAction(
 ): Promise<CouponCodeFormState> {
   'use server';
 
-  const submission = parseWithZod(formData, { schema: couponCodeActionFormDataSchema });
+  const submission = parseWithZod(formData, { schema: couponCodeActionFormDataSchema({}) });
 
   if (submission.status !== 'success') {
     return {
--- a/apps/web/vibes/soul/examples/sections/cart/actions.ts
+++ b/apps/web/vibes/soul/examples/sections/cart/actions.ts
@@ -115,13 +115,15 @@ export async function couponCodeAction(
       const couponCode = submission.value.couponCode;
 
       return {
-        couponCodes: [couponCode],
+        couponCodes: [...prevState.couponCodes, couponCode],
         lastResult: submission.reply({ resetForm: true }),
       };
     }
     case 'delete': {
       return {
-        couponCodes: [],
+        couponCodes: [...prevState.couponCodes].filter(
+          (code) => code !== submission.value.couponCode,
+        ),
         lastResult: submission.reply({ resetForm: true }),
       };
     }
--- a/apps/web/vibes/soul/examples/sections/cart/actions.ts
+++ b/apps/web/vibes/soul/examples/sections/cart/actions.ts
@@ -115,15 +115,13 @@ export async function couponCodeAction(
       const couponCode = submission.value.couponCode;
 
       return {
-        couponCodes: [...prevState.couponCodes, couponCode],
+        couponCodes: [couponCode],
         lastResult: submission.reply({ resetForm: true }),
       };
     }
     case 'delete': {
-      const couponCode = submission.value.couponCode;
-
       return {
-        couponCodes: prevState.couponCodes.filter((code) => code !== couponCode),
+        couponCodes: [],
         lastResult: submission.reply({ resetForm: true }),
       };
     }
--- a/apps/web/vibes/soul/examples/sections/cart/electric.tsx
+++ b/apps/web/vibes/soul/examples/sections/cart/electric.tsx
@@ -161,7 +161,7 @@ async function getCart() {
     ),
   );
 
-  return { ...cart, couponCodeAction: couponCodeAction };
+  return { ...cart, couponCodeAction };
 }
 
 async function setCart(cart: Cart<CartLineItem>) {
--- a/apps/web/vibes/soul/examples/sections/cart/electric.tsx
+++ b/apps/web/vibes/soul/examples/sections/cart/electric.tsx
@@ -8,6 +8,8 @@ import { Button } from '@/vibes/soul/primitives/button';
 import { Cart, CartLineItem } from '@/vibes/soul/sections/cart';
 import { cartLineItemActionFormDataSchema } from '@/vibes/soul/sections/cart/schema';
 
+import { couponCodeAction } from './actions';
+
 export default async function Preview() {
   return (
     <div className="p-4">
@@ -158,7 +160,8 @@ async function getCart() {
         JSON.stringify(generateCartFromLineItems(defaultLineItems)),
     ),
   );
-  return cart;
+
+  return { ...cart, couponCodeAction: couponCodeAction };
 }
 
 async function setCart(cart: Cart<CartLineItem>) {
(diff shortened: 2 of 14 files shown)


apps/web/vibes/soul/form/input/index.tsx
```diff
@@ -27,19 +27,19 @@ import { Label } from '@/vibes/soul/form/label';
  */
 export const Input = React.forwardRef<
   React.ComponentRef<'input'>,
-  Omit<React.ComponentPropsWithoutRef<'input'>, 'id'> & {
+  React.ComponentPropsWithoutRef<'input'> & {
```
jorgemoya, line 30: Needed a way to use this component with a custom label, so I had to pass the `id` generated from conform-to. Curious why we omitted id in the first place.

apps/web/vibes/soul/sections/cart/client.tsx
```diff
@@ -42,6 +43,13 @@ export interface Cart<LineItem extends CartLineItem> {
   summaryItems: CartSummaryItem[];
   total: string;
   totalLabel?: string;
+  couponCodes?: string[];
+  couponCodeAction?: Action<CouponCodeFormState, FormData>;
+  couponCodeCtaLabel?: string;
+  couponCodeDisabled?: boolean;
+  couponCodeLabel?: string;
+  couponCodePlaceholder?: string;
+  couponCodeRemoveLabel?: string;
```
jorgemoya, line 52: I'm not sure if these props need to be part of the `cart` prop, or if they should be root props in the `Cart` section.

apps/web/vibes/soul/sections/cart/coupon-code-form.tsx
```diff
@@ -0,0 +1,109 @@
...
+            key={fields.couponCode.id}
+            placeholder={placeholder}
+          />
+          <SubmitButton disabled={disabled}>{ctaLabel}</SubmitButton>
+        </div>
+        {form.errors?.map((error, index) => <FieldError key={index}>{error}</FieldError>)}
+      </form>
+      <div className="flex flex-wrap gap-1.5 pt-3">
+        {actionStateCouponCodes.map((couponCode) => (
+          <form action={formAction} key={couponCode}>
```
jorgemoya, line 77: Using the same action with `intent`. For these forms, I'm not using `getFormProps`, etc, seems to be only useful for forms with inputs, but I'm not convinced if this is the right way.

apps/web/vibes/soul/examples/sections/cart/actions.ts
```diff
@@ -87,3 +91,41 @@ export async function checkoutAction(
...
+  formData: FormData,
+): Promise<CouponCodeFormState> {
+  'use server';
+
+  const submission = parseWithZod(formData, { schema: couponCodeActionFormDataSchema });
+
+  if (submission.status !== 'success') {
+    return {
+      ...prevState,
+      lastResult: submission.reply({ formErrors: ['This code has expired or is invalid!'] }),
```
migueloller, line 106: This is an example so it doesn't matter much. But when `submission.status` doesn't equal `'success'`, we can just do `submission.reply()` as the error is a parsing error for one of the fields. Submitting a custom form error would be appropriate when the simulated network request fails, for example.

apps/web/vibes/soul/examples/sections/cart/electric.tsx
```diff
@@ -158,7 +160,8 @@ async function getCart() {
         JSON.stringify(generateCartFromLineItems(defaultLineItems)),
     ),
   );
-  return cart;
+
+  return { ...cart, couponCodeAction };
```
migueloller, line 164: This is meant to provide cart data, not props. We should instead pass the `couponCodeAction` prop directly in JSX

apps/web/vibes/soul/form/input/index.tsx
migueloller: I think it's because we were passing hardcode IDs that would conflict. Your approach here is appropriate where we generate one if not provided one. The user of this component just needs to make sure to provide a unique ID as well, which I think conform does

migueloller: One more thing... this component should probably be renamed `InputField` as it's a form field, not just an input. And given that, we probably need to stop spreading props into the `<input>`.

Something to think about @hunterbecton, probably for after we release the Soul VIBE

apps/web/vibes/soul/sections/cart/coupon-code-form.tsx
```diff
@@ -0,0 +1,109 @@
...
+    lastResult,
+    shouldValidate: 'onBlur',
+    shouldRevalidate: 'onInput',
+    onValidate({ formData }) {
+      return parseWithZod(formData, { schema: couponCodeActionFormDataSchema });
+    },
+  });
+
+  return (
+    <div className="border-t border-contrast-100 py-6">
```
andrewreifman, line 56: Need to match the padding in the mockups. Top padding should be 16px, bottom should be 20px.

apps/web/vibes/soul/sections/cart/client.tsx
migueloller: They should be props in the `Cart` section, IMO. These are not properties that you would have in a cart entity in an ecommerce API, right?

Also, I'd nest these inside a `coupon` prop that's optional and use the presence of that prop to determine whether to render the `CouponCodeForm` component instead of using the presence of the action for that.

apps/web/vibes/soul/sections/cart/coupon-code-form.tsx
```diff
@@ -0,0 +1,109 @@
...
+              required: true,
+              type: 'text',
+            })}
+            disabled={disabled}
+            errors={fields.couponCode.errors}
+            id={fields.couponCode.id}
+            key={fields.couponCode.id}
+            placeholder={placeholder}
+          />
+          <SubmitButton disabled={disabled}>{ctaLabel}</SubmitButton>
```
andrewreifman, line 71: The styles of the Submit button look slightly off. The x-padding looks smaller than it should and the text appears lighter in weight. Also, I notice that when the button changes to a loading state, you can't see the loader since the loader color blends in with the white background. We should either disable the hover while it's in the loading state, or change the loader color when hovered.

apps/web/vibes/soul/sections/cart/coupon-code-form.tsx
migueloller: How come we have two forms, though? We should only have 1. If we have 1 then we don't need to use `getFormProps` here

apps/web/vibes/soul/sections/cart/coupon-code-form.tsx
```diff
@@ -0,0 +1,109 @@
...
+        </div>
+        {form.errors?.map((error, index) => <FieldError key={index}>{error}</FieldError>)}
+      </form>
+      <div className="flex flex-wrap gap-1.5 pt-3">
+        {actionStateCouponCodes.map((couponCode) => (
+          <form action={formAction} key={couponCode}>
+            <input name="couponCode" type="hidden" value={couponCode} />
+            <Chip
+              buttonName="intent"
+              buttonValue="delete"
```
migueloller, line 81: I know this isn't part of this PR but it feels weird to not have these just be named `name` and `value`

apps/web/vibes/soul/sections/cart/coupon-code-form.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  return (
+    <div className="border-t border-contrast-100 py-6">
+      <form {...getFormProps(form)} action={formAction} className="space-y-2">
+        <label htmlFor={fields.couponCode.id}>{label}</label>
+        <div className="flex gap-1.5">
+          <Input
+            {...getInputProps(fields.couponCode, {
+              required: true,
+              type: 'text',
```
andrewreifman, line 63: For the empty validation, can you change the text to read "Please enter a valid promo code"

andrewreifman (changes requested):

migueloller (commented):
This is looking great! Got a handful of comments on some form patterns, naming, etc.

migueloller:
@jorgemoya, I think maybe instead of a hidden input perhaps we can use the `formAction` prop together with `.bind` to apply the given coupon's ID?

apps/web/vibes/soul/sections/cart/coupon-code-form.tsx
jorgemoya: Agreed!

apps/web/vibes/soul/sections/cart/client.tsx
jorgemoya: Moving to root.

jorgemoya:
> @jorgemoya, I think maybe instead of a hidden input perhaps we can use the `formAction` prop together with `.bind` to apply the given coupon's ID?

I tried this and was getting some type issues. Reason I have two forms is because each coupon code needs to pass the id. Using `bind` would solve this, let me try again.

apps/web/vibes/soul/sections/cart/coupon-code-form.tsx
jorgemoya: There is an open issue to fix the loader problem.

apps/web/vibes/soul/form/input/index.tsx
jorgemoya: In this case, which is the correct Input to use?

migueloller: I think what you're doing here is right 👍 

No changes needed for this particular comment. My comments re: `InputField`, etc. were more for future changes to VIBES

migueloller:
Sounds good. If you run into issues, let me know. Using multiple forms works too. So don't feel like you need to change your approach if it was working well. It might be nice to use `useForm` in each of these so maybe we extract that "chip" form to its own component, but I'll leave that as an optional thing up to you 🍹

jorgemoya:
> Sounds good. If you run into issues, let me know. Using multiple forms works too. So don't feel like you need to change your approach if it was working well. It might be nice to use `useForm` in each of these so maybe we extract that "chip" form to its own component, but I'll leave that as an optional thing up to you 🍹

Good idea, after playing around with `.bind` it is not working as we expect. I will instead extract the chip form to another component.

apps/web/vibes/soul/sections/cart/schema.ts
```diff
@@ -14,3 +14,14 @@ export const cartLineItemActionFormDataSchema = z.discriminatedUnion('intent', [
     id: z.string(),
   }),
 ]);
+
+export const couponCodeActionFormDataSchema = z.discriminatedUnion('intent', [
+  z.object({
+    intent: z.literal('apply'),
+    couponCode: z.string({ required_error: 'Please enter a valid promo code' }),
```
jorgemoya, line 21: Not sure how we'll be able to translate these 🤔

jorgemoya:
@andrewreifman @migueloller I believe I have addressed all comments. Feel free to re-review!

jorgemoya:
@migueloller I do wonder, since I'm still using multiple forms, if there should be two actions, one for the "apply" and one for "delete".

I think the `intent` pattern works for Cart Line Items because they all have the same `id` associated to the intent, and this id is hidden in all cases.

in my case, either `apply` or `delete` intents, need a distinct id that needs to be passed to the action (one hidden and one as an input), which is why each chip has it's own form.

apps/web/vibes/soul/sections/cart/coupon-code-form.tsx
andrewreifman: Cool, should I go ahead and resolve this convo for now then?

migueloller:
We could definitely do that. Long term I would like to adopt a single pattern in VIBES instead of having both. For this one, do whichever you think is appropriate

apps/web/vibes/soul/sections/cart/schema.ts
migueloller: Make the schema a function that takes the label and returns the schema. I think we've done this for a couple of schemas already.

jorgemoya: True! Updated

jorgemoya:
> We could definitely do that. Long term I would like to adopt a single pattern in VIBES instead of having both. For this one, do whichever you think is appropriate

Will leave as is for now.

hunterbecton (approved):
Looks good! I think the pending state of the coupon button needs work. Right now it looks like the entire button turns white.

migueloller (approved):
This looks great! Nice work @jorgemoya

andrewreifman (approved):

PR #511, opened 2025-02-13
audit:card-carousel

hunterbecton:
- Renamed existing types that used `WithKey` to `WithId`
- Reworked `CarouselEmptyState` to include new CSS variables for styling light and dark color schemes
- Moved to Sections
- Renamed "Components" to "Primitives" in site navigation

I noticed existing components used `id` for more than just passing into the `key` for components. I rather keep the data and naming consistent, so I reverted some data and components that are outside the `CardCarousel`

### DIFF
--- a/apps/web/vibes/soul/sections.ts
+++ b/apps/web/vibes/soul/sections.ts
@@ -26,6 +26,12 @@ export const sections = [
     files: ['sections/blog-post-list/index.tsx'],
   },
   {
+    name: 'card-carousel',
+    dependencies: ['clsx', 'lucide-react'],
+    registryDependencies: ['carousel', 'skeleton', 'card', 'streamable'],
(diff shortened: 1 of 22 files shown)


apps/web/vibes/soul/sections.ts
```diff
@@ -25,6 +25,18 @@ export const sections = [
     registryDependencies: [],
     files: ['sections/blog-post-list/index.tsx'],
   },
+  {
+    name: 'card-carousel',
```
jamesqquick, line 29: I think this is a duplicate right?

jamesqquick: Also, does this depend on skeleton now?

jamesqquick (changes requested):

PR #521, opened 2025-02-21
audit: banner

hunterbecton:
## What/Why
- exports `BannerProps` interface
- reworks CSS to rely on flex box and not absolute positioning for the banner action button
- adds CSS variable for `font-family`
- removed `forwardRef` now that `ref` can be accessed in props


apps/web/vibes/soul/primitives/banner/index.tsx
```diff
@@ -17,71 +26,60 @@ import { ForwardedRef, forwardRef, ReactNode, useCallback, useEffect, useState }
...
-            }}
-          >
-            <X absoluteStrokeWidth size={20} strokeWidth={1.5} />
-          </button>
-        )}
+  return (
+    <div
+      className={clsx(
+        'relative flex items-center justify-between gap-4 overflow-hidden bg-[var(--banner-background,hsl(var(--primary)))] px-8 py-3 transition-all duration-300 ease-in @container',
+        banner.dismissed ? 'pointer-events-none max-h-0' : 'max-h-32',
```
jamesqquick, line 61: After dismissing the banner it still stays. I think setting max height to 0 doesn't account for padding. we might also need to set `py-0`

jamesqquick (changes requested):

apps/web/vibes/soul/primitives/banner/index.tsx
hunterbecton: I moved the padding down to another container because setting padding to 0 would also animate it. Should be good to go now.

jamesqquick (approved):

PR #528, opened 2025-02-24
`Banner` Docs

jamesqquick:
- update example to add a button to reset the banner after it's been dismissed
- added usage section with note about how state is tracked in local storage
- added figma link
- added API reference including CSS Variables


apps/web/vibes/soul/examples/primitives/banner/index.tsx
```diff
@@ -1,10 +1,37 @@
...
+    setHidden(false);
+  }
+
   return (
-    <Banner id="example-banner">
-      Get <strong>15% off</strong> and free shipping with discount code{' '}
-      <strong>&quot;welcome&quot;</strong>
-    </Banner>
+    <>
+      {!hidden && (
```
hunterbecton, line 22: The banner doesn't animate anymore since we're rendering it based on if it's hidden or not. Maybe we should only show the reset button if the value in `localStorage` has the banner hidden? I'm not sure if that would fix the problem.

jamesqquick: Added a delay to acomodate for the animation time before showing the button.

PR #531, opened 2025-02-25
audit: countdown

hunterbecton:
## What/Why
- Move Countdown to Sections in the navigation
- Export `CountdownProps` interface
- Add `twoWeeksFromNow` in the example to ensure there's always a working date
- Add CSS vars

### DIFF
--- a/apps/web/vibes/soul/examples/sections/countdown/index.tsx
+++ b/apps/web/vibes/soul/examples/sections/countdown/index.tsx
@@ -1,11 +1,14 @@
 import { Countdown } from '@/vibes/soul/sections/countdown';
 
 export default function Preview() {
+  const twoWeeksFromNow = new Date();
+  twoWeeksFromNow.setDate(twoWeeksFromNow.getDate() + 14);
+
   return (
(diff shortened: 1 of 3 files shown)


apps/web/vibes/soul/examples/sections/countdown/index.tsx
```diff
@@ -1,10 +1,13 @@
...
 
 export default function Preview() {
+  const twoWeeksFromNow = new Date();
+  twoWeeksFromNow.setDate(twoWeeksFromNow.getDate() + 14);
+
   return (
     <div className="flex flex-col gap-3">
       <Countdown
-        targetDate={new Date('2024-10-24T00:00:00')}
+        targetDate={twoWeeksFromNow}
```
jamesqquick, line 10: I like using two weeks from now, but maybe we need to change the title so it's not misleading since that won't line up with actual Next.js conf date? Should we put something more generic like "Our event is coming"

jamesqquick (changes requested):

jamesqquick (approved):

PR #582, opened 2025-03-09
audit: icon block

hunterbecton:
## What / Why
- adds css vars
- exports props interface
- simplifies some CSS choices like using `border` instead of a weird combo of `divider` and `border`

### DIFF
--- a/apps/web/vibes/soul/sections/icon-block/index.tsx
+++ b/apps/web/vibes/soul/sections/icon-block/index.tsx
@@ -10,25 +10,41 @@ export interface IconBlockProps {
   }>;
 }
 
+/**
+ * This component supports various CSS variables for theming. Here's a comprehensive list, along
+ * with their default values:
+ *
+ * ```css
+ * :root {
+ *   --icon-block-background: hsl(var(--background));
+ *   --icon-block-border: hsl(var(--contrast-100));
+ *   --icon-block-font-family: var(--font-family-body);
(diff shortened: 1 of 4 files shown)


apps/web/vibes/soul/sections/icon-block/index.tsx
```diff
@@ -10,25 +10,41 @@ export interface IconBlockProps {
...
+ *   --icon-block-description: hsl(var(--contrast-400));
+ *   --icon-block-icon: hsl(var(--foreground));
+ * }
+ * ```
+ */
 export const IconBlock = function IconBlock({ list }: IconBlockProps) {
   return (
-    <section className="bg-background text-foreground @container">
-      <ul className="mx-auto flex w-full max-w-screen-2xl flex-wrap justify-center divide-y divide-contrast-100 px-3 @2xl:px-20">
+    <section className="bg-[var(--icon-block-background,hsl(var(--background)))] font-[family-name:var(--icon-block-font-family,var(--font-family-body))] @container">
```
jamesqquick, line 30: If I wrap the `IconBlock` component in a flex container, it doesn't take up any inherent width. Is that expected? I would expect it to continue to take up full width. Maybe it's our use of flex-wrap?

jamesqquick (changes requested):

jamesqquick (approved):

PR #592, opened 2025-03-11
audit: cart

hunterbecton:
## What/Why
- adds CSS vars
- exports props interface
- replaces skeleton with `Skeleton` component


apps/web/vibes/soul/sections/cart/client.tsx
```diff
@@ -276,57 +298,20 @@ function CounterForm({
...
-          >
-            <Plus
-              className="text-contrast-300 transition-colors duration-300 group-hover:text-foreground"
-              size={18}
-              strokeWidth={1.5}
-            />
-          </button>
-        </div>
-
+        <Counter decrementAriaLabel={decrementLabel} incrementAriaLabel={incrementLabel} />
```
jamesqquick, line 302: I think this change is actually breaking functionality. The increment and decrement buttons inside of the form need to have a value associated with them so that we can determine what to do in the action handler. 

In the original code, those buttons had these name and value properties. This allows us to differentiate on the server the difference between incrementing and decrementing.

```
    name="intent"
     value="decrement"
```

This value is then used in the action handler to decide what to do

jamesqquick (changes requested):

PR #605, opened 2025-03-18
site: update homepage for initial launch

hunterbecton:
## What / Why
- 
- adds `currentColor` support for Arrow svg
- adds hover "coming soon" to upcoming VIBES
- adds hover "View docs" to current SOUL vibe
- rewords CTA for newsletter at bottom
- updates images on intro doc and homepage for SOUL


jamesqquick (changes requested):

hunterbecton:
> @hunterbecton I think we'd like to be communicating via email to people more so than when just new VIBES is created. I wonder what @andrewreifman thinks about something like "Stay in the loop for the latest updates from VIBES"?

Yeah, I think that's an excellent callout!

andrewreifman:
Looking at the preview link, I'm wondering if we want to add a navbar to the landing page with mostly the same links from within the docs. Maybe even the search bar? I'm just thinking that it could give more visual weight and people could jump to certain parts of the docs more quickly. Thoughts?

hunterbecton:
> Looking at the preview link, I'm wondering if we want to add a navbar to the landing page with mostly the same links from within the docs. Maybe even the search bar? I'm just thinking that it could give more visual weight and people could jump to certain parts of the docs more quickly. Thoughts?

Just linking to the docs would be fine IMO for V1. I think linking to docs is clear enough, plus it's a pretty common pattern seen on other sites. If we did add links to sections, primitives, etc, I think we would want to change up some of the content on the landing page to speak to some of these things.

hunterbecton:
> Looking at the preview link, I'm wondering if we want to add a navbar to the landing page with mostly the same links from within the docs. Maybe even the search bar? I'm just thinking that it could give more visual weight and people could jump to certain parts of the docs more quickly. Thoughts?

Just linking to the docs would be fine IMO for V1. I think linking to docs is clear enough, plus it's a pretty common pattern seen on other sites. If we did add links to sections, primitives, etc, I think we would want to change up some of the content on the landing page to speak to some of these things.

andrewreifman (approved):

PR #609, opened 2025-03-19
feat: Add Switch form component

jordanarldt:
- Add `Switch` component to vibes

image


apps/web/vibes/soul/form/switch/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+ * ```css
+ *  :root {
+ *    --switch-primary-background: hsl(var(--primary));
+ *    --switch-secondary-background: hsl(var(--foreground));
+ *    --switch-tertiary-background: hsl(var(--background));
+ *    --switch-tertiary-border: hsl(var(--contrast-200));
+ *  }
+ * ```
+ */
+export const Switch = ({
```
jorgemoya, line 35: Should there be a `className` prop?

PR #611, opened 2025-03-19
feat: Add Tooltip primitive

jordanarldt:
Add `Tooltip` primitive component

image


apps/web/vibes/soul/primitives/tooltip/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  delayDuration?: number;
+  skipDelayDuration?: number;
+  trigger?: React.ReactNode;
+  open?: boolean;
+  setOpen?: (open: boolean) => void;
+  side?: 'top' | 'right' | 'bottom' | 'left';
+  sideOffset?: number;
+}
+
+// eslint-disable-next-line valid-jsdoc
```
jorgemoya, line 16: Should we remove this eslint-disable in VIBES?

apps/web/vibes/soul/primitives/tooltip/index.tsx
```diff
@@ -0,0 +1,62 @@
+import * as TooltipPrimitive from '@radix-ui/react-tooltip';
+import { clsx } from 'clsx';
+
+interface TooltipProps extends React.PropsWithChildren {
+  align?: 'center' | 'end' | 'start';
+  className?: string;
+  delayDuration?: number;
+  skipDelayDuration?: number;
+  trigger?: React.ReactNode;
```
jorgemoya, line 9: In what scenarios would we want a trigger being optional for tooltip?

apps/web/vibes/soul/primitives/tooltip/index.tsx
jordanarldt: Ope yep! Good call

apps/web/vibes/soul/primitives/tooltip/index.tsx
jordanarldt: If it's being controlled externally by some other state or if the trigger needs to be detached from the component. For our use-case in catalyst it doesn't need to be optional, so I can make it required if needed!

But also, the trigger isn't required in the underlying Radix UI component IIRC!

PR #617, opened 2025-03-20
fix(Streamable): preserve inputs identity if none of them is a promise

agurtovoy:
Backporting https://github.com/bigcommerce/catalyst/pull/2116 with a test.


apps/web/vibes/soul/lib/streamable.tsx
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/consistent-indexed-object-style */
```
agurtovoy, line 1: We removed this rule, no need to suppress it.

fikrikarim (approved):

PR #625, opened 2025-03-26
chore: upgrade next and next-mdx-remote

apledger:


hunterbecton (commented):
I reviewed this and think getting off canary and rc would be nice. I'm not 100% clear why we needed the `react` and `react-dom` rc's though, but didn't run across any issues testing. Do you recall why we needed to use these branches? I want to make sure I'm testing any features that these branches support that the upgraded ones might not.

hunterbecton (approved):

PR #644, opened 2025-04-10
Hunter/vib 4122 update installation page in vibes

hunterbecton:
## What / Why
- fixes duplicate fonts being imported into example with `removeDuplicatesByName` function
- makes `getLayoutCode` function more verbose so it's easier to follow and read
- removes `// [!code highlight]` from examples because it was being copied onto the clipboard ( more details below )
- rewords instructions to be clearer
- adds option in `tailwind.config` file for those using `src` directory in their Next.js files

### removes `// [!code highlight]` from examples because it was being copied onto the clipboard

Currently, when you copy the code from examples that include highlighted lines, the transformer `// [!code highlight]` is also included. I tried to fix this by getting the code using a `ref` and `innerText`, but was having issues with the line breaks disappearing. I decided the easiest option for now is to just remove `// [!code highlight]` from the examples and try and fix this later.

## Testing
https://github.com/user-attachments/assets/6c70f1f1-d2eb-434b-a412-52ad30c8c060

### DIFF
--- a/apps/web/components/brand/brand-installation.tsx
+++ b/apps/web/components/brand/brand-installation.tsx
@@ -28,13 +28,6 @@ interface Props {
   };
 }
 
-function highlightLines(str: string) {
-  return str
-    .split('\n')
-    .map((line) => line.concat('// [!code highlight]'))
-    .join('\n');
-}
-
 function getVariableCode(cssVars: CSSVars) {
   return `:root {\n${Object.entries(cssVars)
     .filter(([name]) => !name.startsWith('--font-family'))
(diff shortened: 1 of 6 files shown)


apps/web/components/brand/brand-installation.tsx
```diff
@@ -29,10 +29,7 @@ interface Props {
 }
 
 function highlightLines(str: string) {
-  return str
-    .split('\n')
-    .map((line) => line.concat('// [!code highlight]'))
-    .join('\n');
+  return str.split('\n').join('\n');
```
apledger, line 32: This is a noop?

apledger (dismissed):
LGTM

apps/web/vibes/soul/docs/installation.mdx
```diff
@@ -4,16 +4,24 @@ title: Installation
 
 <Steps>
   <Step>
-    ### Add dependencies
+    ###  Follow the [Next.js documentation](https://v3.tailwindcss.com/docs/guides/nextjs) to create a new project
```
jamesqquick, line 7:
```suggestion
    ###  Follow the [Tailwind documentation](https://v3.tailwindcss.com/docs/guides/nextjs) to create a new Next.js project using Tailwind 3.
```

jamesqquick (approved):

apps/web/components/brand/brand-installation.tsx
hunterbecton: I reverted this back to the original code. I think I was trying to remove the code highlighting from the fonts section of the code because there was issues highlighting the part of the code with the `body` tag. I'm going to leave it for now and set an [issue in Linear](https://linear.app/makeswift/issue/VIB-4224/code-highlight-not-parsed-correctly).

PR #647, opened 2025-04-17
fix: adjust spacing for price label skeleton

jorgemoya:
Adjust spacing to prevent layout shift when the price loads.


jorgemoya:
Not a perfect solution, still needs to be fixed.

PR #650, opened 2025-04-24
Hunter/vib 4207 create reveal component for soul

hunterbecton:
## What / Why
- adds `Reveal` component and docs
- simplifies previous `AnimatedLink` with `AnimatedUnderline`, which now exports an animated `span`
- creates new docs for `AnimatedUnderline`
- replaces instances of old `AnimatedUnderline`


migueloller (dismissed):
LGTM 👍🏻

PR #672, opened 2025-06-18
fix: overflow issue with shiki code blocks

hunterbecton:
## What / Why
- fixes overflow/scroll issue by setting `overflow-auto` to `<code>` and  removing unnecessary padding, margin, and width from Shiki lines
- fixes issue with one-line code snippets having a higher height than necessary by setting the Shiki container to `flex`

### Display `flex` fix

https://github.com/user-attachments/assets/10dce1d3-21c3-4a35-a865-7ed934516505

### Overflow fix

https://github.com/user-attachments/assets/3a21dff5-c73e-4c1b-ae0c-67c822d3386c

### DIFF
--- a/apps/web/app/(marketing)/style/shiki.css
+++ b/apps/web/app/(marketing)/style/shiki.css
@@ -56,7 +56,7 @@
 }
 
 .shiki {
-  @apply !bg-contrast-100 !rounded-none p-4;
+  @apply !bg-contrast-100 flex !rounded-none p-4;
 }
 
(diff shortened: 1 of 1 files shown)


hunterbecton:
> Looks like an improvement at the least. I would love it if the position of the scrollbars were improved as well, so they were never hidden or feeling to far from the edges of the code example box. I think we can fix that at another time though.

I'll add an issue to track that feedback in Linear.

PR #675, opened 2025-06-25
Add `description` to `BlogPost` type

jamesqquick:
`BlogPost` type has  a `content` property that is currently being used in `BlogPostContent` and assumed to be HTML. However, it is used as a regular string in `BlogPostCard`. This PR adds an optional `description` to `BlogPost` type that can be used in `BlogPostCard` if it's available. Otherwise, it will continue to use the `content` prop.


hunterbecton (commented):
Thanks for adding this! I think we need to add the `description` to the docs and this will be good to go.

PR #676, opened 2025-06-26
feat: add file dropzone component

hunterbecton:
## What/Why
- adds `FileInput` component
- adds docs for `FileInput` component
- adds changelog for Soul
- updates `conform-to/react` to version `^1.8.1` so we could use the new `useControl` hook
- fixes errors where `defaultValue` was not set on `useForm`

## Testing
https://github.com/user-attachments/assets/cd9881fe-7f9f-4cb8-9d16-739dff433d72

### DIFF
--- a/apps/web/vibes/soul/examples/form/file-input/index.tsx
+++ b/apps/web/vibes/soul/examples/form/file-input/index.tsx
@@ -1,0 +1,63 @@
+'use client';
+
+import { useCallback } from 'react';
+
+import { FileInput } from '@/vibes/soul/form/file-input';
+import type { FileInputProps } from '@/vibes/soul/form/file-input';
+
+export default function Preview() {
+  const onUpload: NonNullable<FileInputProps['onUpload']> = useCallback(
+    async (uploadFiles, { onProgress, onSuccess, onError }) => {
+      try {
+        const uploadPromises = uploadFiles.map(async (file, index) => {
+          try {
+            if (index === 1) {
+              await new Promise((resolve) => setTimeout(resolve, 500));
+              onError(file, new Error('Upload failed'));
+              return;
+            }
+            // Simulate file upload with progress
+            const totalChunks = 10;
+            let uploadedChunks = 0;
+
+            // Simulate chunk upload with delays
+            for (let i = 0; i < totalChunks; i = i + 1) {
+              // Simulate network delay (100-300ms per chunk)
+              await new Promise((resolve) => setTimeout(resolve, Math.random() * 200 + 100));
+
+              // Update progress for this specific file
+              uploadedChunks = uploadedChunks + 1;
+              const progress = (uploadedChunks / totalChunks) * 100;
+              onProgress(file, progress);
+            }
+
+            // Simulate server processing delay
+            await new Promise((resolve) => setTimeout(resolve, 500));
+            onSuccess(file);
+          } catch (error) {
+            onError(file, error instanceof Error ? error : new Error('Upload failed'));
+          }
+        });
+
+        await Promise.all(uploadPromises);
+      } catch (error) {
+        console.error('Unexpected error during upload:', error);
+      }
+    },
+    [],
+  );
+
+  return (
+    <div className="p-10">
+      <FileInput
+        accept="image/*"
+        label="Feature Images"
+        maxFiles={2}
+        maxSize={200 * 1024}
+        message="Max 2 files, up to 200KB"
+        multiple
+        onUpload={onUpload}
+      />
+    </div>
+  );
+}
--- a/apps/web/vibes/soul/form/file-input/file-dropzone.tsx
+++ b/apps/web/vibes/soul/form/file-input/file-dropzone.tsx
@@ -1,0 +1,149 @@
+'use client';
+
+import { clsx } from 'clsx';
+import { useCallback, useState } from 'react';
+import type {
+  ClipboardEvent,
+  ComponentPropsWithoutRef,
+  DragEvent,
+  KeyboardEvent,
+  ReactNode,
+  RefObject,
+} from 'react';
+
+export interface FileDropzoneProps extends ComponentPropsWithoutRef<'div'> {
+  children?: ReactNode;
+  inputRef: RefObject<HTMLInputElement | null>;
+  disabled?: boolean;
+  invalid?: boolean;
+}
+
+export function FileDropzone({
+  id,
+  children,
+  inputRef,
+  disabled,
+  invalid,
+  onDrop,
+}: FileDropzoneProps) {
+  const [isDragging, setIsDragging] = useState(false);
+
+  const handleOnDrop = useCallback(
+    (event: DragEvent<HTMLDivElement>) => {
+      setIsDragging(false);
+      onDrop?.(event);
+    },
+    [onDrop],
+  );
+
+  const handleOnDragOver = useCallback((event: DragEvent<HTMLDivElement>) => {
+    event.preventDefault();
+    setIsDragging(true);
+  }, []);
+
+  const handleOnDragEnter = useCallback((event: DragEvent<HTMLDivElement>) => {
+    event.preventDefault();
+    setIsDragging(true);
+  }, []);
+
+  const handleOnDragLeave = useCallback((event: DragEvent<HTMLDivElement>) => {
+    const relatedTarget = event.relatedTarget;
+
+    if (
+      relatedTarget &&
+      relatedTarget instanceof Node &&
+      event.currentTarget.contains(relatedTarget)
+    ) {
+      return;
+    }
+
+    event.preventDefault();
+    setIsDragging(false);
+  }, []);
+
+  const handleOnKeyDown = useCallback(
+    (event: KeyboardEvent<HTMLDivElement>) => {
+      if (event.key === 'Enter' || event.key === ' ') {
+        event.preventDefault();
+        inputRef.current?.click();
+      }
+    },
+    [inputRef],
+  );
+
+  const handleOnPaste = useCallback(
+    (event: ClipboardEvent<HTMLDivElement>) => {
+      event.preventDefault();
+      setIsDragging(false);
+
+      const items = event.clipboardData.items;
+      if (items.length === 0) return;
+
+      const pastedFiles: File[] = [];
+
+      for (const item of items) {
+        if (item.kind === 'file') {
+          const file = item.getAsFile();
+          if (file) {
+            pastedFiles.push(file);
+          }
+        }
+      }
+
+      if (pastedFiles.length === 0) return;
+
+      const inputElement = inputRef.current;
+
+      if (!inputElement) return;
+
+      const dataTransfer = new DataTransfer();
+
+      for (const file of pastedFiles) {
+        dataTransfer.items.add(file);
+      }
+
+      inputElement.files = dataTransfer.files;
+      inputElement.dispatchEvent(new Event('change', { bubbles: true }));
+    },
+    [inputRef],
+  );
+
+  const handleOnClick = useCallback(() => {
+    inputRef.current?.click();
+  }, [inputRef]);
+
+  return (
+    <div
+      aria-controls={`${id}-input ${id}-list`}
+      aria-disabled={disabled}
+      className={clsx(
+        'flex min-h-19 cursor-pointer flex-col items-center justify-center gap-1.5 rounded-lg border-[1.5px] border-dashed border-(--file-input-dropzone-border,var(--contrast-200)) bg-(--file-input-dropzone-background,var(--background)) p-4 transition-colors duration-300 ease-in-out select-none @sm:flex-row',
+        // Hover styles
+        'hover:border-(--file-input-dropzone-border-hover,var(--foreground)) hover:bg-(--file-input-dropzone-background-hover,var(--contrast-100))',
+        // Focus styles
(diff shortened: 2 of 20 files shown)


jamesqquick (dismissed):

apps/web/vibes/soul/form/file-upload/index.tsx
```diff
@@ -0,0 +1,1244 @@
...
+        data-invalid={itemContext.fileState?.status === 'error' ? '' : undefined}
+        id={id}
+        role="listitem"
+      >
+        <FileUploadItemMetadata />
+        <FileUploadItemDelete />
+        <FileUploadItemProgress />
+        <span className="sr-only" id={statusId}>
+          {statusText}
+        </span>
```
bc-micah, line 1068: This text seems to be generated within the component and I don't see a mechanism for the user of the component to control these, I see that the docs mention internationalization support out of the box for vibes - am I missing something?

bc-micah (commented):
I am not sure I understand the need for setting up a custom store, the layers of the component seem to be simple enough to not need that level of complexity. Not sure if we considered using a simpler approach like useReducer and context.

I leave that to you but it'd be good to document the reason in the PR or the comment.

As a consumer of this component in B2B, my main concern is the internationalization aspect - seems like there are status / error strings that are used internal to the component that is not updatable from the outside. I am not sure if I am missing something there.

Also there are no tests to speak of, and considering we added our own state management - I would expect to see some level of testing. Is that not an expectation on vibes components as I do not see anything setup around that?

apps/web/vibes/soul/form/file-upload/index.tsx
```diff
@@ -0,0 +1,1244 @@
...
+      </div>
+    </FileUploadItemContext.Provider>
+  );
+};
+
+const FileUploadItemMetadata = ({ children }: ComponentPropsWithoutRef<'div'>) => {
```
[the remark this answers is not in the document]
hunterbecton: Just an old pattern I forgot to remove–thanks for catching it!

apps/web/vibes/soul/form/file-upload/index.tsx
```diff
@@ -0,0 +1,1244 @@
...
+
+  return (
+    <div
+      className="fade-in-0 slide-in-from-top-2 animate-in mt-3 grid gap-3"
+      id={context.listId}
+      role="list"
+    >
+      {files.map((file, index) => (
+        <FileUploadItem
+          key={`${file.name}-${file.size}-${file.lastModified}-${index}`}
```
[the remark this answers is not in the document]
hunterbecton: I think that makes sense. I know [React docs ](https://react.dev/reference/react/useId#caveats) say to not use `useId` to generate keys. Would `crypto.randomUUID()` be safe to use here?

migueloller:
re: translating labels, +1 from on making sure that we expose labels so that they can be translated.

re: custom store, it does seem this could be powered by `useReducer` without us having to set up our own listeners, etc. The rest of the store logic seems like it would fit in nicely with `useReducer`. Curious if you tried that out @hunterbecton 

re: tests, also +1 from me there. Thing is, we don't have a test suite for VIBES right now and I feel like adding it to this PR might make it a lot bigger than it currently is, and it's already quite big. @hunterbecton, thoughts on prioritizing a small project to add tests to the more complex parts of VIBES?

One more thing, I do think that a 1,200+ line file makes it very hard to follow what's going on. I would appreciate if we broke down this module into some smaller ones both to tell a better "story" in this PR (it's so big) and also make it easier for the consumer of the code to understand it.

hunterbecton:
Thanks for the feedback, @migueloller. I can look into simplifying the component by removing the custom store and splitting it into more sub-components. 

I know I've mentioned the need for testing in the past, and I would love to prioritize this after the CLI project is shipped. I was also interested in using Storybook 9 for this project. I think I mentioned this to you recently and you seemed on board with it. Of course, I'd still think about it some more before jumping into using it.

hunterbecton:
I refactored this `FileUpload` component not to use a custom store but instead use the `useReducer` hook. I also broke the component into the following files:
- `FileUploadProvider`
- `FileUploadDropzone`
- `FileUploadList`
- `FileUploadItem`

My intent was to put most of the logic into the `FileUploadProvider` so it's easier to parse the other related components.

I also realized that the lint errors that I was getting from Conform after updating were from not passing in a default value into the `useForm` hook.

apps/web/vibes/soul/form/file-upload/file-upload-dropzone.tsx
```diff
@@ -0,0 +1,189 @@
...
+      const dataTransfer = new DataTransfer();
+      for (const file of filesInStore) {
+        dataTransfer.items.add(file);
+      }
+
+      hiddenInputRef.current.files = dataTransfer.files;
+    };
+
+    syncFiles();
+  }, [state, hiddenInputRef]);
```
migueloller, line 91: This effect won't re-run when the hidden input ref changes because it's a ref. Given that we conditionally hide the hidden input based on `register`, if that prop changes then we will not get the latest value. I think that using React state for the ref should address this. Then, instead of `register` we can use refs and can expose it via `useImperativeHandle`

migueloller, line 91: What does this hook do? Is it synchronizing the state with the DOM (i.e., file input)? It might be helpful to extract this to a custom hook and give it some description with JSDoc so that it's clear what it's doing.

That said, is there a reason we didn't derive this state from the input directly instead of tracking it in React state?

apps/web/vibes/soul/form/file-upload/file-upload-dropzone.tsx
```diff
@@ -0,0 +1,189 @@
...
+            accept={accept}
+            disabled={disabled}
+            hidden
+            multiple={multiple}
+            ref={setHiddenInputRef}
+            required={required}
+            type="file"
+            {...registerProps}
+          />
+        )}
```
migueloller, line 177: How come we have a hidden input in addition to the regular input? 🤔

apps/web/vibes/soul/form/file-upload/file-upload-item.tsx
```diff
@@ -0,0 +1,209 @@
...
+  idleLabel = 'Ready to upload',
+  errorLabel = 'Error',
+}: FileUploadItemProps) {
+  const id = useId();
+  const statusId = `${id}-status`;
+  const nameId = `${id}-name`;
+  const sizeId = `${id}-size`;
+  const messageId = `${id}-message`;
+
+  const { state, onRemoveFile } = useFileUpload();
```
migueloller, line 39: IMO, we should name this `removeFile` and leave names like `onRemoveFile` to a prop to add an event listener

apps/web/vibes/soul/form/file-upload/file-upload-item.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+  const fileState = state.files.get(file);
+  const fileCount = state.files.size;
+  const fileIndex = Array.from(state.files.keys()).indexOf(file) + 1;
+
+  let statusText: string;
+
+  if (Boolean(fileState?.error)) {
```
migueloller, line 47: Instead of coercing, can we do `fileState?.error != null`?

apps/web/vibes/soul/form/file-upload/file-upload-provider.tsx
migueloller, line 1: Haven't taken a look at this file just yet. Wanted to provide some feedback on consuming the provider API before doing so.

apps/web/vibes/soul/form/file-upload/file-upload-dropzone.tsx
hunterbecton: After our conversation last night, I'll update this to not have this weird pattern and decouple it from Conform.

apps/web/vibes/soul/form/file-upload/index.tsx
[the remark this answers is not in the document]
hunterbecton: No, what were you thinking?

apps/web/vibes/soul/form/dynamic-form/index.tsx
```diff
@@ -376,5 +377,8 @@ function DynamicFormField({
           selected={typeof controls.value === 'string' ? new Date(controls.value) : undefined}
         />
       );
+
+    case 'file-upload':
+      return <></>;
```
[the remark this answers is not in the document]
hunterbecton: Yes, I'll separate the dynamic form into a separate PR.

apps/web/vibes/soul/primitives/alert/index.tsx
```diff
@@ -44,7 +44,7 @@ export function Alert({
   return (
     <div
       className={clsx(
-        'flex max-w-[356px] min-w-[284px] items-center justify-between gap-2 rounded-xl border border-(--alert-border,color-mix(in_oklab,var(--foreground)_10%,transparent)) py-3 ps-4 pe-3 shadow-xs',
+        'flex max-w-[356px] items-center justify-between gap-2 rounded-xl border border-(--alert-border,color-mix(in_oklab,var(--foreground)_10%,transparent)) py-3 ps-4 pe-3 shadow-xs',
```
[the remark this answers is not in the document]
hunterbecton: Yeah, I noticed that alerts could have different widths depending on the content inside, so setting the Sonner to `w-full` and removing the `min-width` here made sure there were no alerts with widths between 284-356px.

apps/web/vibes/soul/form/file-upload/index.tsx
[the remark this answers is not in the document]
hunterbecton: I like that suggestion.

hunterbecton:
I took the feedback from @migueloller and decoupled the `FileInput` from Conform. This will make `FileInput` more straightforward to implement with Conform or other form libraries that a user might want to use, as it relies on the native file input element. 

I will create a separate PR to implement the `FileInput` into the `DynamicForm` component.

apps/web/vibes/soul/examples/form/file-upload/electric.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+        // Wait for all uploads to complete
+        await Promise.all(uploadPromises);
+      } catch (error) {
+        // This handles any error that might occur outside the individual upload processes
+        console.error('Unexpected error during upload:', error);
+      }
+    },
+    [],
+  );
```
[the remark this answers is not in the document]
hunterbecton: I realized we didn't need the different example files since the content doesn't change, therefore, reusing this function should be fixed.

apps/web/vibes/soul/form/file-upload/file-upload-item.tsx
icatalina: curious here, what's the problem with the coercion? I find the single = more troublesome... 🤔

apps/web/vibes/soul/examples/form/file-input/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+        await Promise.all(uploadPromises);
+      } catch (error) {
+        console.error('Unexpected error during upload:', error);
+      }
+    },
+    [],
+  );
+
+  return (
+    <>
```
joshuawootonn, line 51: extra fragment

apps/web/vibes/soul/form/file-input/file-dropzone.tsx
```diff
@@ -0,0 +1,323 @@
...
+      if (inputRef.current) {
+        inputRef.current.value = '';
+      }
+    },
+    [onFilesAdded],
+  );
+
+  return (
+    <>
+      <Label className={clsx(hideLabel === true ? 'sr-only' : 'mb-2')} htmlFor={labelId}>
```
joshuawootonn, line 242: super nit: generally when a value is explicitly a boolean, even if it's optional, we don't compare it to an explicit value.

apps/web/vibes/soul/form/file-input/file-dropzone.tsx
```diff
@@ -0,0 +1,323 @@
...
+          // Disabled styles
+          'data-[disabled]:cursor-not-allowed data-[disabled]:border-(--file-input-dropzone-border-disabled,color-mix(in_oklab,var(--contrast-200)_70%,transparent)) data-[disabled]:bg-(--file-input-dropzone-background-disabled,var(--background)) data-[disabled]:focus-visible:ring-0 data-[disabled]:focus-visible:ring-offset-0 data-[disabled]:focus-visible:outline-none',
+          // Invalid styles
+          'data-[invalid]:border-(--file-input-dropzone-border-error,var(--error))',
+          // Dragging styles
+          'data-[dragging]:border-(--file-input-dropzone-border-dragging,var(--foreground)) data-[dragging]:bg-(--file-input-dropzone-background-dragging,var(--success-highlight))',
+        )}
+        data-disabled={disabled === true ? '' : undefined}
+        data-dragging={isDragging ? '' : undefined}
+        data-invalid={invalid === true ? '' : undefined}
```
joshuawootonn, line 263: I like the fact that you broke up tailwind styles by pseudo state. I've started doing that and I think it's helpful!

joshuawootonn (commented):
From my context this is looking good 👍🏼

migueloller (approved):
This is looking a lot better! I've got a couple more comments and there's also some of the patterns we discussed in person to better separate concerns. I'll leave it up to you @hunterbecton whether you want to pursue these now or not.

PR #677, opened 2025-07-20
Build public registry for Soul components to be installed via CLI

hunterbecton:
## What / Why
This PR enables the easy installation of VIBE components and all their dependencies using the VIBES CLI. This will significantly enhance the development experience, as the current manual installation process can be tedious, especially for components with many dependencies. 

- scaffolds start of `vibes` CLI
- adds a script that builds a public component registry for Soul components
- adds `tsx` dev dep for running build script
- updates docs to include CLI installation steps for each component
- adds `build:next` and `build:registry` commands to `package.json` scripts
- alphabetizes components in registry files
- moves registry files into `registry` folder
- moves form components to `form` folder
- adds missing related files and deps
- removes components that no longer exist from registry files

## Test

https://github.com/user-attachments/assets/818cf997-c046-4efb-bc97-66a0fb3d0edd


apps/web/vibes/soul/registry/primitives.ts
```diff
@@ -121,32 +101,6 @@ export const primitives = [
     registryDependencies: ['streamable', 'skeleton'],
     files: ['primitives/cursor-pagination/index.tsx'],
   },
-  {
-    name: 'date-picker',
-    dependencies: ['lucide-react', '@radix-ui/react-popover'],
-    registryDependencies: ['input', 'calendar'],
-    files: ['form/date-picker/index.tsx'],
-  },
```
joshuawootonn, line 129: Was this file the default shadcn references? 

Trying to understand why there were so many removals. If ^ is the case then makes sense 👍🏼

apps/web/.gitignore
```diff
@@ -34,3 +34,6 @@ yarn-error.log*
 # typescript
 *.tsbuildinfo
 next-env.d.ts
+
+# Generated registry files
+public/r/
```
joshuawootonn, line 39: Missing new line

apps/cli/src/lib/registry.ts
```diff
@@ -0,0 +1,53 @@
...
+  registryDependencies?: string[];
+  files: Array<{
+    path: string;
+    content: string;
+    type: string;
+    target?: string;
+  }>;
+}
+
+const DEFAULT_REGISTRY_URL = "https://localhost:3000/r";
```
joshuawootonn, line 16: Will this be changed before this thing is shipped? Or is it meant to be hit a local runtime?

joshuawootonn (commented):
This is very cool ⚡

apps/cli/src/lib/registry.ts
hunterbecton: > Will this be changed before this thing is shipped? Or is it meant to be hit a local runtime?

Yes, it needs to be changed!

> I don't like the `/r` pathname btw. I would prefer this is explicit.

Works for me 👍

apps/web/vibes/soul/registry/primitives.ts
hunterbecton: The files in this folder are used to generate the docs and to populate the registry. There are so many removals because we had form components in the primitives file, and it made more sense to me to separate these into their own file since we separate them in the docs.

Also, it threw several errors when I ran the build script for files that did not exist anymore, so I removed those definitions.

PR #704, opened 2025-08-15
feat: Add Table component

davelinke:
Adding the table component requested by the B2B effort


hunterbecton:
Thanks for this PR! Some errors need to be addressed above with the deployment, but I still took a look and had some thoughts. 

We've designed VIBE components to follow a single-component pattern over a compound-component pattern. Although the compound pattern is popular with libraries like shadcn, we wanted to make VIBE components handle that for you. This means that as a developer, all you have to do is pass in the data to a single `Table` component—no need to nesting components.

To get here, we probably need to think more about what the data looks like that's coming into the `Table` component, so we know how to render it.

